PR #13, opened 2020-06-22
Fixing of resize operations on Cert outputs

abi87:


src/gtest/test_checktransaction.cpp
```diff
@@ -510,5 +510,296 @@ TEST(checktransaction_tests, ScCertVersion) {
...
+
+TEST(CertificateManipulation, NonEmptyCertTransformationToMutableIsReversible) {
+    //create non-empty transaction
+    CMutableScCertificate helperMutCert;
+    unsigned int OutNum = 10;
+    for(unsigned int idx = 0; idx < OutNum; ++idx)
+        helperMutCert.addOut(CTxOut(CAmount(idx),CScript()));
+
+    unsigned int bwtOut = 3;
+    CScript bwtScript = CScript() << OP_DUP << OP_HASH160 << ToByteVector(uint160()) << OP_EQUALVERIFY << OP_CHECKSIG;
```
abi87, line 639: @alsala  I am currently forced to create a dummy bwtScript since, in subsequent transformation from mutable cert to cert, GetHash() is called and a transformation to CBackwardTransfer is carried out; this fails if script is empty. Should instead map an empty CTxOut to an empty CBackwardTransfer?

src/gtest/test_checktransaction.cpp
```diff
@@ -510,5 +510,296 @@ TEST(checktransaction_tests, ScCertVersion) {
...
+    CTransaction        revertedTxByCopyCtor(mutByCopyCtor);
+
+    EXPECT_FALSE(EmptyOriginalTx == revertedTxByCopyCtor);
+    EXPECT_TRUE(EmptyOriginalTx.GetHash().IsNull());
+    EXPECT_FALSE(revertedTxByCopyCtor.GetHash().IsNull());
+
+    // AssignOp -> CopyCtor
+    CMutableTransaction mutByAssignOp;
+    mutByAssignOp = EmptyOriginalTx;
+    CTransaction revertedTxFromAssignement(mutByCopyCtor);
```
alsala, line 526: Shouldn't be mutByAssignOp  instead of mutByCopyCtor? Or why it has been instantiated?

src/gtest/test_checktransaction.cpp
```diff
@@ -510,5 +510,296 @@ TEST(checktransaction_tests, ScCertVersion) {
...
+    CMutableTransaction mutByCopyCtor(nonEmptyOriginalTx);
+    CTransaction        revertedTxByCopyCtor(mutByCopyCtor);
+
+    EXPECT_TRUE(nonEmptyOriginalTx == revertedTxByCopyCtor)
+        <<" nonEmptyOriginalTx.GetHash() "<<nonEmptyOriginalTx.GetHash().ToString()
+        <<" revertedTxByCopyCtor.GetHash() "<<revertedTxByCopyCtor.GetHash().ToString();
+
+    // AssignOp -> CopyCtor
+    CMutableTransaction mutByAssignOp;
+    mutByAssignOp = nonEmptyOriginalTx;
```
alsala, line 560: Same as above

src/gtest/test_checktransaction.cpp
alsala: Yes, I agree

src/gtest/test_checktransaction.cpp
```diff
@@ -510,5 +510,296 @@ TEST(checktransaction_tests, ScCertVersion) {
...
+TEST(CertificateManipulation, NonEmptyCertTransformationToMutableIsReversible) {
+    //create non-empty transaction
+    CMutableScCertificate helperMutCert;
+    unsigned int OutNum = 10;
+    for(unsigned int idx = 0; idx < OutNum; ++idx)
+        helperMutCert.addOut(CTxOut(CAmount(idx),CScript()));
+
+    unsigned int bwtOut = 3;
+    CScript bwtScript = CScript() << OP_DUP << OP_HASH160 << ToByteVector(uint160()) << OP_EQUALVERIFY << OP_CHECKSIG;
+    for(unsigned int idx = 0; idx < OutNum; ++idx)
```
alsala, line 640: probably OutNum is a typo and bwtOut should be used here

src/gtest/test_checktransaction.cpp
abi87: You are right, it should be mutByAssignOp

src/gtest/test_checktransaction.cpp
abi87: yes

src/gtest/test_checktransaction.cpp
abi87: yes, indeed

PR #14, opened 2020-06-25
Rework wallet objects hierarchy

abi87:
Removed need for slicing CWalletTx/Cert. Introduced UTs for wallet

### DIFF
--- a/src/wallet/gtest/test_wallet_cert.cpp
+++ b/src/wallet/gtest/test_wallet_cert.cpp
@@ -9,7 +9,7 @@
 
 class CertInWalletTest : public ::testing::Test {
 public:
-    CertInWalletTest(): walletDbLocation(), wallet("wallet.dat"), pWalletDb(nullptr) {}
+    CertInWalletTest(): walletDbLocation(), pWallet(nullptr), pWalletDb(nullptr) {}
     ~CertInWalletTest() = default;
 
     void SetUp() override {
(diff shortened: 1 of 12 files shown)


src/wallet/gtest/test_wallet_cert.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+class CertInWalletTest : public ::testing::Test {
+public:
+    CertInWalletTest(): walletDbLocation(), pWallet(nullptr), pWalletDb(nullptr) {}
+    ~CertInWalletTest() = default;
+
+    void SetUp() override {
+        //Setup environment
+        SelectParams(CBaseChainParams::TESTNET);
+        boost::filesystem::path walletDbLocation = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
```
alsala, line 18: use a different name, just to disambiguate with data member and improve readability

src/wallet/gtest/test_wallet_cert.cpp
```diff
@@ -0,0 +1,124 @@
...
+        pWallet = new CWallet("wallet.dat");
+        try {  pWalletDb = new CWalletDB(pWallet->strWalletFile, "cr+"); }
+        catch(std::exception& e) {
+            ASSERT_TRUE(false)<<"Could not create tmp wallet db for reason "<<e.what();
+        }
+    };
+
+    void TearDown() override {
+        std::vector<std::shared_ptr<CWalletTransactionBase> > vWtx;
+        DBErrors nZapWalletRet = pWallet->ZapWalletTx(vWtx);
```
alsala, line 32: nZapWalletRet is not used, should it be checked?

src/zcbenchmarks.cpp
```diff
@@ -291,11 +291,11 @@ double benchmark_try_decrypt_notes(size_t nAddrs)
     }
 
     auto sk = libzcash::SpendingKey::random();
-    auto tx = GetValidReceive(*pzcashParams, sk, 10, true);
+    auto walletTx = GetValidReceive(*pzcashParams, sk, 10, true);
 
     struct timeval tv_start;
     timer_start(tv_start);
-    auto nd = wallet.FindMyNotes(tx);
+    auto nd = wallet.FindMyNotes(walletTx.getWrappedTx());
```
alsala, line 298: nd does not seem to be used? (It was the same in the original code anyway)

src/wallet/gtest/test_wallet_cert.cpp
abi87: the `ZapWalletTx `here makes sure that the wallet is duly cleaned-up and different UTs are well decoupled.
I would check `ZapWalletTx `return type with an EXPECT_TRUE(), in order to signal potential coupling of UTs, should `ZapWalletTx `fail.

src/wallet/gtest/test_wallet_cert.cpp
abi87: Actually I intended yo use the data member, removing declaration here

src/zcbenchmarks.cpp
abi87: It's a benchmark, so I believe they want to verify the time it takes for finding notes

PR #16, opened 2020-06-26
As/py tests

alsala:
Move some common api to test framework, added minor fixes and code cleanup, added short comments for summarizing tests scope


qa/rpc-tests/sc_cert_invalidate.py
```diff
@@ -51,8 +51,24 @@ def disconnect_nodes(self, from_connection, node_num):
         while any(peer['version'] == 0 for peer in from_connection.getpeerinfo()):
             time.sleep(0.1)
 
+    def gen_block(self, sc_info, scid, nIdx = 0 ):
```
abi87, line 54: I would rename this to refresh_sidechain, because the block generation seems instrumental to the update of the sidechain state

PR #20, opened 2020-07-02
New sbh wallet rpc cmd + fixed rescan wallet

abi87:


src/wallet/gtest/test_wallet_cert.cpp
```diff
@@ -0,0 +1,1250 @@
...
+    //Add block information
+    CWalletTx walletTx(pWallet, transparentTx);
+    CBlock txBlock;
+    txBlock.vtx.push_back(transparentTx);
+    walletTx.hashBlock = txBlock.GetHash();
+    walletTx.SetMerkleBranch(txBlock);
+    walletTx.fMerkleVerified = true; //shortcut
+
+    chainSettingUtils::ExtendChainActiveWithBlock(txBlock);
+    int txCreationHeight = chainActive.Height();
```
alsala, line 238: txCreationHeight is not used

abi87: Done

src/wallet/rpcwallet.cpp
```diff
@@ -847,7 +847,10 @@ UniValue create_sidechain(const UniValue& params, bool fHelp)
             "                                          hexadecimal format. Used as public input for WCert proof verification. Its size must be " + strprintf("%d", SC_FIELD_SIZE) + " bytes\n"
             "}\n"
             "\nResult:\n"
-            "\"transactionid\"    (string) The resulting transaction id.\n"
+            "{\n"
```
i-Alex, line 903: What about `sc_create` command output? We should change it similar to current one.
Do we need to keep both commands?

alsala: Ok agreed.
For the time being I would keep both versions of the command.

src/wallet/rpcwallet.cpp
```diff
@@ -81,12 +81,104 @@ void EnsureWalletIsUnlocked()
...
+            int bwtMaturityDepth = sidechain.StartHeightForEpoch(currentEpoch+1) +
+                sidechain.SafeguardMargin() - nHeight;
+            bwtMaturityHeight = bwtMaturityDepth + nHeight;
+        } else {
+            // if conf < 0 we can not tell
+        }
+    }
 
     UniValue vout(UniValue::VARR);
     for (unsigned int i = 0; i < tx.getTxBase()->GetVout().size(); i++) {
```
albertog78, line 184: valueZat -> valueSat

PR #21, opened 2020-07-03
As/leftover code reviews

alsala:
Fix of several findings left from the previous code review, such as removal of visitor pattern in sc related rpc cmds, removal of default param fron some api and more modifications in mempool flow


abi87:
@alsala , as agreed AFK, I have added the UT and fix for GetSidechain on mempool. Please review it

abi87:
Absorbed into https://github.com/HorizenOfficial/zend_oo/pull/75 since cmds were to be updated to introduce mainchain backward transfer requests

PR #30, opened 2020-07-13
Fix to removeImmatureExpenditures method in mempool; added a UT for i…

alsala:
…t; fixed a py script after branch rebase


src/gtest/test_sidechain_to_mempool.cpp
```diff
@@ -449,6 +450,49 @@ TEST_F(SidechainsInMempoolTestSuite, FwdsAndCertInMempool_FwtRemovalDoesNotAffec
     EXPECT_TRUE(mempool.mapSidechains.at(scId).backwardCertificate == cert.GetHash());
 }
 
+TEST_F(SidechainsInMempoolTestSuite, ImmatureExpenditureRemoval) {
+    //Create and mature coinbase
```
abi87, line 454: Edit this to `Create coinbase`, since maturing has a different comment below

alsala: done

PR #37, opened 2020-07-21
As/issue 36

alsala:
getrawtransaction and decoderawtransaction RPC cmds can handle both tx and cert


src/main.cpp
```diff
@@ -1655,6 +1656,100 @@ bool GetCertificate(const uint256 &hash, CScCertificate &certOut, uint256 &hashB
 }
 
 
+bool GetTxBaseObj(const uint256 &hash, std::unique_ptr<CTransactionBase>& pTxBase, uint256 &hashBlock, bool fAllowSlow)
```
abi87, line 1659: The implementation of `GetTxBaseObj` looks like you basically search if it is a tx and if you do not find it you try looking for certificate.
Can't you just make a call to `GetTransaction` and if it return false, try `GetCertificate`? This would avoid the code duplication

src/rpc/rawtransaction.cpp
```diff
@@ -349,19 +349,33 @@ UniValue getrawtransaction(const UniValue& params, bool fHelp)
     if (params.size() > 1)
         fVerbose = (params[1].get_int() != 0);
 
-    CTransaction tx;
+    // allocated by the callee
+    std::unique_ptr<CTransactionBase> pTxBase;
+    
     uint256 hashBlock;
-    if (!GetTransaction(hash, tx, hashBlock, true))
+    if (!GetTxBaseObj(hash, pTxBase, hashBlock, true) || !pTxBase)
```
abi87, line 356: Can `GetTxBaseObj` return true and `pTxBase` be null? as far as I undestand it cannot happen, give current implementation.

alsala: Given the current implementation this should never happen, but the caller should  always check the pTxBase anyway

src/main.cpp
alsala: If we stick to this implementation we might as well remove GetCertificate/GetTransaction, but if we choose to opt for the proposed approach we will leave it. As far as RPC cmds, we will probably remove getrawcertificate and decodedrawcertificate after a further check with client applications.

abi87 (approved):

albertog78 (approved):

PR #39, opened 2020-07-23
Fixup mac compilation

abi87:
We verified the link warning were already present in master and content ourselves with fixing the compilation errors with the minimal amount of Makefile and code changes


abi87:
Absorbed into https://github.com/HorizenOfficial/zend_oo/pull/75, dropping the `-Wno-inconsistent-missing-override` flag since issue has been fixed by introducing override where required

PR #51, opened 2020-09-11
Introducing Poseidon Hash for ScCommitmentTree Verification

abi87:
In the following I have refactored the code verifying the scTxsCommitment tree and replace it with Posidon hash facilities as introduced by new version of zendoo.
While I targeted sidechain_dev, these changes are not complete yet, mainly for two reasons:

1. zendoo is still under active development, which may affect current changes. Given the refactoring, "blast radius" should be minimized
2. error handling in deserializing/serializing into zendoo is to be improved.
3. we may consider introducing a fork to preserve block already validated by previous hash version


src/gtest/test_libzendoo.cpp
```diff
@@ -58,61 +61,82 @@ TEST(ZendooLib, PoseidonHashTest)
     auto rhs_field = zendoo_deserialize_field(rhs);
     ASSERT_TRUE(rhs_field != NULL);
 
+
     auto expected_hash = zendoo_deserialize_field(hash);
     ASSERT_TRUE(expected_hash != NULL);
 
-    const field_t* hash_input[] = {lhs_field, rhs_field};
+    auto digest = ZendooPoseidonHash(NULL, 0);
```
DanieleDiBenedetto, line 68: This constructor is a little bit ugly when no personalization is needed. I will add an empty one too in zendoo-mc-cryptolib.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+        field_t* scIdField = mapCertToField(scid);
+
+        std::vector<field_t*> singleSidechainComponents = {ftRootField, btrRootField, certRootField, scIdField};
+        field_t * sidechainTreeRoot = merkleTreeRootOf(singleSidechainComponents);
+
+        vSortedScLeaves.push_back(sidechainTreeRoot);
+    }
+
+    field_t * finalTreeRoot = emptyField.zero;
```
DanieleDiBenedetto, line 95: Probably it's best to actually compute the root even if all leaves are zero (or set it to a phantom root, or the actual hash of all 0 nodes up until the root). Keep in mind always the "SNARK friendliness" we want to achieve

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+    uint256 res = mapFieldToHash(finalTreeRoot);
+    if (finalTreeRoot != emptyField.zero) zendoo_field_free(finalTreeRoot);
+
+    return res;
+}
+#endif
+
+field_t* SidechainTxsCommitmentBuilder::mapScTxToField(const uint256& ccoutHash, const uint256& txHash, unsigned int outPos) const
+{
+    static_assert((sizeof(uint256) + sizeof(uint256) + sizeof(unsigned int)) < SC_FIELD_SIZE);
```
DanieleDiBenedetto, line 108: It's not enough to be < SC_FIELD_SIZE. With the MNTs, for instance, a size that is always safe to deserialize is 752 bits, i.e. 94 bytes. Probably I can introduce a similar constant to call from zendoo-mc-cryptolib, or you just omit this assert (zendoo_deserialize_field will throw error anyway if you try to deserialize something above the field modulus)

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+    memcpy(hash, toHash.append(SC_FIELD_SIZE - toHash.size(), '\0').c_str(), SC_FIELD_SIZE);
+
+    //generate field element. MISSING null-check
+    field_t* field = zendoo_deserialize_field(hash);
+    return field;
+}
+
+field_t* SidechainTxsCommitmentBuilder::mapCertToField(const uint256& certHash) const
+{
+    static_assert(sizeof(uint256) < SC_FIELD_SIZE);
```
DanieleDiBenedetto, line 124: Same comment as above

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+        throw;
+
+    unsigned char field_bytes[SC_FIELD_SIZE];
+    zendoo_serialize_field(pField, field_bytes);
+
+    uint256 res = Hash(BEGIN(field_bytes), END(field_bytes));
+    return res;
+}
+
+unsigned int SidechainTxsCommitmentBuilder::treeHeightForLeaves(unsigned int numberOfLeaves) const
```
DanieleDiBenedetto, line 146: It's ok for a temporary implementation, but in the SNARKs we need to have constant height for all the Merkle Trees, so this function won't be needed.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+{
+    //Notes: leaves are consumed, i.e. freed and nulled, if non-empty
+    //       root is a deep-copy, hence it must be freed later on
+    //       null leaves are mapped to emptyField
+    //       it is guaranteed not to return nullptr
+    field_t* res = nullptr;
+    auto btrTree = ZendooGingerRandomAccessMerkleTree(treeHeightForLeaves(leaves.size()));
+    for(field_t* & leaf: leaves)
+    {
+        //As of now we swallow possible errors in deserialization of field by replacing null with empty field
```
DanieleDiBenedetto, line 165: This is to be discussed. Are genuine deserialization error supposed to happen ? A malicious user is able to craft a valid transaction that would lead to such error ? If yes, do we skip the leaf or do we set it to zero ?

src/sc/sidechainTxsCommitmentBuilder.h
```diff
@@ -0,0 +1,62 @@
...
+    std::map<uint256, std::vector<field_t*> > mScMerkleTreeLeavesBtr;
+
+    // Key:   the side chain ID
+    // Value: the field into with the certificate related to this scid is mapped
+    std::map<uint256, field_t* > mScCerts;
+
+    // the catalog of scid, resulting after having collected all above contributions
+    std::set<uint256> sScIds;
+
+    class EmptyField
```
DanieleDiBenedetto, line 42: It's not necessary true that an empty leaf/node, corresponds to "all zeros" (also looking at the implementation in the .cpp). I think that this class must depend on the height of the node for which we are computing the empty field (i.e. an empty leaf will have some default value (not necessarily all zeros), an empty node at level 1 will be the hash of two default leaves, ..., an empty node at level i will be the hash of two empty nodes at level i-1).
Of course all this values can be precomputed and used when necessary (I can easily provide a function in zendoo-mc-cryptolib)

DanieleDiBenedetto (commented):
Some general comments here and there, mostly related to SNARK friendliness, that need further discussion

src/gtest/test_libzendoo.cpp
```diff
@@ -126,20 +150,20 @@ TEST(ZendooLib, TestProof)
 
     //Inputs
     unsigned char end_epoch_mc_b_hash[32] = {
-        78, 85, 161, 67, 167, 192, 185, 56, 133, 49, 134, 253, 133, 165, 182, 80, 152, 93, 203, 77, 165, 13, 67, 0, 64,
-        200, 185, 46, 93, 135, 238, 70
+        157, 219, 85, 159, 75, 56, 146, 21, 107, 239, 76, 31, 208, 213, 230, 24, 44, 74, 250, 66, 71, 23, 106, 4, 138,
+        157, 28, 43, 158, 39, 152, 91
     };
 
     unsigned char prev_end_epoch_mc_b_hash[32] = {
```
alsala, line 157: It would be better to avoid hardcoding the value 32 (similarly to what has been done with SC_FIELD_SIZE)

src/sc/sidechainTxsCommitmentBuilder.cpp
alsala, line 108: the syntax of static_assert used here compiles only from c++17 on, c+11 wants an error msg too:

static_assert ( bool_constexpr , message )    (since C++11)
static_assert ( bool_constexpr  )                    (since C++17)

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    return res;
+}
+#endif
+
+field_t* SidechainTxsCommitmentBuilder::mapScTxToField(const uint256& ccoutHash, const uint256& txHash, unsigned int outPos) const
+{
+    static_assert((sizeof(uint256) + sizeof(uint256) + sizeof(unsigned int)) < SC_FIELD_SIZE);
+
+    std::string toHash = std::string(ccoutHash.begin(), ccoutHash.end()) +
```
alsala, line 110: avoid using std::string for concatenating bytes array, since if a 0x00 is contained its size is wrongly calculated since the underlying string is terminated at that byte.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+void SidechainTxsCommitmentBuilder::add(const CTransaction& tx)
+{
+    if (!tx.IsScVersion())
+        return;
+
+    unsigned int ccOutIdx = 0;
+    LogPrint("sc", "%s():%d -getting leaves for vsc out\n", __func__, __LINE__);
+    for(; ccOutIdx < tx.GetVscCcOut().size(); ++ccOutIdx)
+    {
+        uint256 scId = tx.GetVscCcOut().at(ccOutIdx).GetScId();
```
alsala, line 30: In order to avoid useless copy, use can be made of const uint256&.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+        std::vector<field_t*>& vec = mScMerkleTreeLeavesFt[scId]; //create or pick entry for scId
+        field_t* pField = mapScTxToField(tx.GetVscCcOut()[ccOutIdx].GetHash(), tx.GetHash(), ccOutIdx);
+        vec.push_back(pField);
+    }
+
+    LogPrint("sc", "%s():%d -getting leaves for vft out\n", __func__, __LINE__);
+    for(; ccOutIdx < tx.GetVftCcOut().size(); ++ccOutIdx)
+    {
+        uint256 scId = tx.GetVftCcOut().at(ccOutIdx).GetScId();
```
alsala, line 41: In order to avoid useless copy, use can be made of const uint256&.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+        field_t* pField = mapScTxToField(tx.GetVftCcOut()[ccOutIdx].GetHash(), tx.GetHash(), ccOutIdx);
+        vec.push_back(pField);
+    }
+
+    LogPrint("sc", "%s():%d - nIdx[%d]\n", __func__, __LINE__, ccOutIdx);
+}
+
+void SidechainTxsCommitmentBuilder::add(const CScCertificate& cert)
+{
+    uint256 scId = cert.GetScId();
```
alsala, line 54: In order to avoid useless copy, use can be made of const uint256&.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+        field_t * sidechainTreeRoot = merkleTreeRootOf(singleSidechainComponents);
+
+        vSortedScLeaves.push_back(sidechainTreeRoot);
+    }
+
+    field_t * finalTreeRoot = emptyField.zero;
+    if (vSortedScLeaves.size() != 0)
+        finalTreeRoot = merkleTreeRootOf(vSortedScLeaves);
+
+    uint256 res = mapFieldToHash(finalTreeRoot);
```
alsala, line 99: In order to avoid useless copy, use can be made of const uint256&.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+uint256 SidechainTxsCommitmentBuilder::mapFieldToHash(const field_t* pField) const
+{
+    if (pField == nullptr)
+        throw;
+
+    unsigned char field_bytes[SC_FIELD_SIZE];
+    zendoo_serialize_field(pField, field_bytes);
+
+    uint256 res = Hash(BEGIN(field_bytes), END(field_bytes));
```
alsala, line 142: In order to avoid useless copy, use can be made of const uint256&.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,161 @@
...
+        auto itCert = mScCerts.find(scid);
+        if (itCert != mScCerts.end() )
+        {
+            certRootField = itCert->second;
+        }
+
+        field_t* scIdField = mapCertToField(scid);
+
+        std::vector<field_t*> singleSidechainComponents = {ftRootField, btrRootField, certRootField, scIdField};
+        field_t * sidechainTreeRoot = merkleTreeRootOf(singleSidechainComponents);
```
i-Alex, line 79: Current solution is different to the previous one. Previously we calculated at first the merkle tree root for `{ftRootField, btrRootField}` -> `txsHash`. Then, we did it once more for `{txsHash, certRootField, scIdField}`
Honestly, I don't remember exactly the reason, but seems that such order was important (see page 17 in zendoo paper).

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,161 @@
...
+    //       It is guaranteed not to return nullptr.
+
+    unsigned int numberOfLeaves = leaves.size();
+    if (numberOfLeaves == 0) 
+        numberOfLeaves = 1;
+    
+    auto btrTree = ZendooGingerMerkleTree(treeHeightForLeaves(numberOfLeaves), numberOfLeaves);
+    for(field_t* & leaf: leaves)
+    {
+        if (leaf != nullptr) {
```
i-Alex, line 151: 1. previously we had a magic string hash in case of empty leaf. does current way is safe?
2. What if there is no leaves at all at some level (for example no FT and BTRs)? What is the behavior?

i-Alex (changes requested):

src/sc/sidechainTxsCommitmentBuilder.cpp
DanieleDiBenedetto: 1. For the moment we have assumed that if a leaf is nullptr, we just skip appending it. It's a design decision, it can be discussed (remember that this is still a draft, lot of things must change, like the fact that we will have trees of fixed height, while here we are instantiating trees of a height dependent on the number of leaves each time);

2. MerkleTree implementation in zendoo-mc-cryptolib (GingerLib) already handles the case in which you append no leaf, for a tree of whatever height. The value it returns will be a root actually computed like normal assuming that all the leaves are a "zero" field element (or some field element value you set to represent an empty leaf).

i-Alex: Cool! thanks for clarifications.

src/sc/sidechainTxsCommitmentBuilder.cpp
abi87: as discussed AFK, gcc will use return value optimization even with -O0, and will not use it only if the flag  -fno-elide-constructors is specified. If you agree, I'd avoid the reference construct.

src/sc/sidechainTxsCommitmentBuilder.cpp
abi87: as discussed AFK, gcc will use return value optimization even with -O0, and will not use it only if the flag  -fno-elide-constructors is specified. If you agree, I'd avoid the reference construct.

src/sc/sidechainTxsCommitmentBuilder.cpp
abi87: as discussed AFK, gcc will use return value optimization even with -O0, and will not use it only if the flag  -fno-elide-constructors is specified. If you agree, I'd avoid the reference construct.

src/sc/sidechainTxsCommitmentBuilder.cpp
abi87: as discussed AFK, gcc will use return value optimization even with -O0, and will not use it only if the flag  -fno-elide-constructors is specified. If you agree, I'd avoid the reference construct.

src/sc/sidechainTxsCommitmentBuilder.cpp
abi87: as discussed AFK, gcc will use return value optimization even with -O0, and will not use it only if the flag  -fno-elide-constructors is specified. If you agree, I'd avoid the reference construct.

src/sc/sidechainTxsCommitmentBuilder.cpp
abi87: Indeed I changed wrt to previous implementation. I made it slightly closer to what @DanieleDiBenedetto told me the final implementation would be (with fix number of leaves for fwds, btrs and certs and number of sidechains too).
Bear in mind that this is not yet an "active" pull request, and I made this only to have early feedbacks and ease up the following changes to our crypto libs.
Moreover, since current validations are only in a separate testnet, we may even have no need to ensure backward compatibility.

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,161 @@
...
+        uint256 scId = tx.GetVscCcOut().at(ccOutIdx).GetScId();
+        sScIds.insert(scId);
+
+        std::vector<field_t*>& vec = mScMerkleTreeLeavesFt[scId]; //create or pick entry for scId
+        field_t* pField = mapScTxToField(tx.GetVscCcOut()[ccOutIdx].GetHash(), tx.GetHash(), ccOutIdx);
+        vec.push_back(pField);
+    }
+
+    LogPrint("sc", "%s():%d -getting leaves for vft out\n", __func__, __LINE__);
+    for(; ccOutIdx < tx.GetVftCcOut().size(); ++ccOutIdx)
```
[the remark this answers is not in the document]
abi87: Nice catch, I got tricked by the refactoring I did. Fixed and extended a UT to capture the case

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+    return static_cast<unsigned int>(ceil(log2f(static_cast<float>(numberOfLeaves))));
+}
+
+field_t* SidechainTxsCommitmentBuilder::merkleTreeRootOf(std::vector<field_t*>& leaves) const
+{
+    //Notes: Leaves are consumed, i.e. freed and nulled;
+    //       It is guaranteed not to return nullptr.
+
+    unsigned int numberOfLeaves = leaves.size();
```
i-Alex, line 186: `numberOfLeaves` computation doesn't consider `null` leaves. So in case of null leaves, we will put less entries, than expected in `btrTree`.

abi87: @i-Alex , thanks for your observation. 
You are correct. Null leaves come out only of error in serialization. Current draft just swallows this error and goes on with trying and building the tree. This will crash upon encountering null leaves.
We will introduce a different handling of serialization, making the add fail in case null leave is returned, so to be able to assert that all leaves submitted to the tree are not null.
Moreover the final tree must have a fixed number of leaves (and a fixed height consequently) which we need yet to determine (I guess give limits on block size and minimal sizes of sidechain related transactions)

i-Alex: Just want to add the case with `btrRootField` that is always null for now and it's correct (maybe):
```
std::vector<field_t*> singleSidechainComponents = {ftRootField, btrRootField, certRootField, scIdField};
field_t * sidechainTreeRoot = merkleTreeRootOf(singleSidechainComponents);
```

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -0,0 +1,203 @@
...
+        field_t* scIdField = mapCertToField(scid);
+
+        std::vector<field_t*> singleSidechainComponents = {ftRootField, btrRootField, certRootField, scIdField};
+        field_t * sidechainTreeRoot = merkleTreeRootOf(singleSidechainComponents);
+
+        vSortedScLeaves.push_back(sidechainTreeRoot);
+    }
+
+    field_t * finalTreeRoot = merkleTreeRootOf(vSortedScLeaves);
+    uint256 res = mapFieldToHash(finalTreeRoot);
```
i-Alex, line 83: Seems, that we need to return exactly poseidon hash, which will lead to the changes in the Block structure.
@DanieleDiBenedetto In the sidechain we have a problem now: `MerklePath.validate(...)` expects the root hash (poseidon one) as one of the arguments, but, we have only the sha hash of that root hash.

Temporary this can be solved by introducing `MerklePath.apply(leaf)` method that will return poseidon root hash. Then it can be hashed again with sha and compared with the on in MC block header. But, seems, we need to avoid sha at all.

abi87:
PR superseeded by https://github.com/HorizenOfficial/zend_oo/pull/83 where also a better wrapper for field_t ptr has been introduced (CFieldElement)

PR #55, opened 2020-10-08
Speedup SelectCoins with remove/erase idioms

abi87:
Scanning a vector and removing specific items one by one can cause relocation and results in a O(N^2) complexity.
[Erase/Remove idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom)  guarantees linear complexity instead since all elements marked for removal are erased at once


abi87:
Absorbed by https://github.com/HorizenOfficial/zend_oo/pull/74

PR #58, opened 2020-11-12
Refactored mempool::remove to extract list of dependencies

abi87:
As some corner case from multiple certificate handling, we need the list of dependencies for a given tx/cert in mempool.
I have extracted a function providing just this info from mempool::remove which, in its recursive case, eliminates from mempool a given tx/cert and the full list of its mempool dependencies.
I think the refactoring does improves also readability of mempool::remove which performs a depth-first-seach on the mempool direct acyclic graph of txes and cert dependencies.

Targeting our base branch for quality cert development


alsala (approved):

abi87:
Closing down since this has been picked up in PR https://github.com/HorizenOfficial/zend_oo/pull/59 along with improvements from code review

PR #59, opened 2020-11-23
Cert quality handling

abi87:


src/txmempool.cpp
```diff
@@ -1150,3 +1227,169 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+            const uint256& memPoolCertHash = entry.second;
+            const CScCertificate& memPoolCert = mempool.mapCertificate.at(memPoolCertHash).GetCertificate();
+
+            // should be nothing but the same epoch
+            if (memPoolCert.epochNumber != cert.epochNumber)
+            {
+                LogPrint("cert", "%s:%s():%d - WARNING: cert %s epoch=%d vs cert %s epoch=%d\n", __FILE__,__func__, __LINE__,
+                    cert.GetHash().ToString(), cert.epochNumber, memPoolCertHash.ToString(), memPoolCert.epochNumber);
+                continue;
+            }
```
abi87, line 1255: If this situation happens we have messed up badly the mempool I believe. We should add checks on CTxMempool::check() to make sure this does not happen (or we assert)

src/txmempool.cpp
```diff
@@ -1150,3 +1227,169 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+                    LogPrint("cert", "%s:%s():%d - cert %s q=%d, fee=%s NOK; q=%d, fee=%s for same sc/epoch is in mempool\n",
+                        __FILE__, __func__, __LINE__, cert.GetHash().ToString(), cert.quality, FormatMoney(certFee),
+                        memPoolCert.quality, FormatMoney(memPoolCertFee));
+                    return false;
+                }
+            }
+        }
+    }
+    LogPrint("cert", "%s:%s():%d - cert %s q=%d OK\n", __FILE__,  __func__, __LINE__, cert.GetHash().ToString(), cert.quality);
+    return true;
```
abi87, line 1278: So if I understand right, this function checks for quality/fee relationship only, but not for vicious spending dependencies?  Maybe it is the right thing to do, but please add comment about it

src/txmempool.cpp
```diff
@@ -1150,3 +1227,169 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
           cachedInnerUsage);
 }
 
+bool CCoinsViewMemPool::IsQualityValid(const CScCertificate& cert, CAmount certFee) const
```
abi87, line 1230: As mentioned in coins.cpp file, I would rename it like `CheckQuality(cert)` or `HaveQuality(cert)`

src/txmempool.cpp
abi87: An alternative would be to skip also the check on fees here and to it later.
We would have a cert for which `checkQuality` is true as long as quality is sufficiently high.
Later on, when checking which, of the possible two same quality certificates to keep in mempool, we would look at fees and spending dependencies.
This function could be renamed like `IsQualityHighEnough(cert)` and you would have a much cleaner interface, avoiding the useless CAmount in CCoinsViewCache::IsQualityValid

abi87 (commented):
Started looking at coins.h/cpp and txmempool.h/cpp and I have written down some comments

src/coins.cpp
```diff
@@ -659,6 +665,27 @@ bool CCoinsViewCache::GetSidechain(const uint256 & scId, CSidechain& targetScInf
     return false;
 }
 
+int64_t CCoinsViewCache::GetTopQualityCert(const uint256& scId, int epochNumber, uint256& hash) const
+{
+    LogPrint("mempool", "%s.%s():%d - sc %s, epoch %d\n", __FILE__, __func__, __LINE__, scId.ToString(), epochNumber);
+
+    int64_t topQual = base->GetTopQualityCert(scId, epochNumber, hash);
```
abi87, line 672: Since GetTopQualityCert is not implemented for CCoinsViewDB, this call is essentially an expensive constructor of type int64_t topQual = CCoinsView::GetTopQualityCert, i.e. int64_t topQual = CScCertificate::QUALITY_NULL;
Just drop this, it's useless, I believe

src/coins.cpp
```diff
@@ -659,6 +665,27 @@ bool CCoinsViewCache::GetSidechain(const uint256 & scId, CSidechain& targetScInf
...
+    LogPrint("mempool", "%s.%s():%d - base: cert [%s], q=%d\n", __FILE__, __func__, __LINE__, hash.ToString(), topQual);
+
+    CSidechain targetInfo;
+    if (GetSidechain(scId, targetInfo)                             &&
+        !targetInfo.lastCertificateHash.IsNull()                   &&
+        epochNumber == targetInfo.lastEpochReferencedByCertificate &&
```
abi87, line 678: I wonder if the check on lastEpochReferencesByCertificate is of any value.
Can't `GetTopQualityCert` return the highest quality for all certificates received for a given sidechain?
With the check on epochNumber, a GetTopQualityCert call after the sidechain is expired would return -1, I believe. Why you need this behaviour?

src/coins.cpp
```diff
@@ -1027,27 +1099,166 @@ bool CCoinsViewCache::UpdateScInfo(const CScCertificate& cert, CTxUndo& certUndo
...
 
+        }
+        else
+        {
+            // should never happen if certs are ordered by quality in a block
+            LogPrint("cert", "%s():%d - cert quality %d not greater than last seen %d", 
+                __func__, __LINE__, cert.quality, scIt->second.scInfo.lastCertificateQuality);
+            return false;
+        }
+    }
```
abi87, line 1178: As discussed offline, this case is temporary, before enforcing certificates ordering in blocks by quality both in consensus and in block formation.
However I believe hard coding such ordering in UpdateScInfo would make in unusable in mempool. I believe this is very bad, since in next future we would want to extend UpdateScInfo checks on CTxMempool::check()

qa/rpc-tests/quality_mempool.py
```diff
@@ -0,0 +1,519 @@
...
+            quality, constant_1, [pkh_node3], [bwt_amount])
+        try:
+            cert_3_epoch_0 = self.nodes[2].send_certificate(scid_1, epoch_number_1, quality, epoch_block_hash_1, normal_quality_proof, amount_cert_3, HIGH_CERT_FEE)
+            mark_logs("Certificate is {}".format(cert_3_epoch_0), self.nodes, DEBUG_MODE)
+        except JSONRPCException, e:
+            errorString = e.error['message']
+            mark_logs("Send certificate failed with reason {}".format(errorString), self.nodes, DEBUG_MODE)
+            assert (False)
+
+        # Checking all certificate in mempool, they are not there because:
```
alsala, line 390: Comments should be updated

qa/rpc-tests/quality_mempool.py
```diff
@@ -0,0 +1,519 @@
...
+        # cert3 has better quality ---> cert1 evicted
+        # cert2 depends on cert1 (spends its change) --> cert2 evicted
+        # cert3 depends on cert2 (spends its change) --> cert3 refused
+        self.sync_all()
+        assert_equal(False, cert_1_epoch_0 in self.nodes[0].getrawmempool())
+        assert_equal(True, cert_2_epoch_0 in self.nodes[0].getrawmempool())
+        assert_equal(True, cert_3_epoch_0 in self.nodes[0].getrawmempool())
+
+        # as a consequence, mempools are misaligned, node1 still has them
```
alsala, line 399: Comments to be removed here

qa/rpc-tests/quality_mempool.py
```diff
@@ -0,0 +1,519 @@
...
+        except JSONRPCException, e:
+            errorString = e.error['message']
+            mark_logs("Send certificate failed with reason {}".format(errorString), self.nodes, DEBUG_MODE)
+            assert(False)
+
+        self.sync_all()
+        assert_equal(True, cert_1_sc1 in self.nodes[0].getrawmempool())
+        assert_equal(True, cert_2_sc2 in self.nodes[0].getrawmempool())
+
+        # Create Cert3 with quality 100 and place it in mempool
```
alsala, line 483: Comment to be updated, quality 150

alsala (changes requested):

qa/rpc-tests/quality_voiding.py
```diff
@@ -0,0 +1,362 @@
...
+
+        # Check balance
+        bwt_address = self.nodes[1].getrawcertificate(cert_3_epoch_0, 1)['vout'][1]['scriptPubKey']['addresses'][0]
+        res = self.nodes[1].listtxesbyaddress(bwt_address)
+        assert_equal(res[0]['scid'], scid)
+        mark_logs("Height : {}".format(res[0]['vout'][1]['maturityHeight']), self.nodes, DEBUG_MODE)
+        ud = self.nodes[1].getunconfirmedtxdata(bwt_address)
+        assert_equal(bwt_amount_2, ud['bwtImmatureOutput'])
+        mark_logs("Immature output : {}".format(ud['bwtImmatureOutput']), self.nodes, DEBUG_MODE)
+
```
alsala, line 360: The test could be extended in this way:
1) node 0 reconsiders block block_inv
2) next maturity height is reached for bwt_amount_3, and voiding of relevant certificates is checked also via the cmd listunspent() (see for instance script sc_cert_change.py)
3) stop and restart of the network and check cert voiding again

qa/rpc-tests/quality_voiding.py
```diff
@@ -0,0 +1,362 @@
...
+        self.sync_all()
+        for i in range(1, NUMB_OF_NODES):
+            assert_equal(sorted(self.nodes[0].getrawmempool()), sorted(self.nodes[i].getrawmempool()))
+
+        mark_logs("Check cert is in mempools", self.nodes, DEBUG_MODE)
+        assert_equal(True, cert_1_epoch_0 in self.nodes[0].getrawmempool())
+
+        # Create Cert2 with lower quality and place it in mempool
+        quality = 90
+        pkh_node2 = self.nodes[1].getnewaddress("", True)
```
alsala, line 249: for consistency this should be called pkh_node1 since belongs to node 1 wallet

qa/rpc-tests/quality_voiding.py
```diff
@@ -0,0 +1,362 @@
...
+        assert_true(cert_2_epoch_0 in self.nodes[0].getblock(self.nodes[0].getbestblockhash(), True)['cert'])
+        assert_false(cert_3_epoch_0 in self.nodes[0].getblock(self.nodes[0].getbestblockhash(), True)['cert'])
+
+        # Check balance
+        bwt_address = self.nodes[1].getrawcertificate(cert_3_epoch_0, 1)['vout'][1]['scriptPubKey']['addresses'][0]
+        res = self.nodes[1].listtxesbyaddress(bwt_address)
+        assert_equal(res[0]['scid'], scid)
+        mark_logs("Height : {}".format(res[0]['vout'][1]['maturityHeight']), self.nodes, DEBUG_MODE)
+        ud = self.nodes[1].getunconfirmedtxdata(bwt_address)
+        assert_equal(bwt_amount_2, ud['bwtImmatureOutput'])
```
alsala, line 358: This assert will fail because it is node 0 that invalidated the block; from node 1 perspective, nothing is changed and the topo quality cert is still cert3.

qa/rpc-tests/quality_voiding.py
```diff
@@ -0,0 +1,362 @@
...
+
+        height = self.nodes[0].getblockcount()
+        mark_logs("Check certificates in chain", self.nodes, DEBUG_MODE)
+        assert_true(cert_1_epoch_0 in self.nodes[0].getblock(self.nodes[0].getblockhash(height - 1), True)['cert'])
+        assert_true(cert_2_epoch_0 in self.nodes[0].getblock(self.nodes[0].getblockhash(height - 1), True)['cert'])
+        assert_true(cert_3_epoch_0 in self.nodes[0].getblock(mined, True)['cert'])
+
+        mark_logs("Checking certificates in mempool after block revertion(via invalidateblock)", self.nodes, DEBUG_MODE)
+        block_inv = self.nodes[0].getbestblockhash()
+        self.nodes[0].invalidateblock(block_inv)
```
alsala, line 338: It would be more interesting to have node 1 invalidating the block (see also comment below) because the bwts of the certs have been sent to its wallet

alsala (changes requested):

qa/rpc-tests/quality_nodes.py
```diff
@@ -0,0 +1,357 @@
...
+
+        self.split_network()
+
+        # Create Cert3 with quality 100 and place it in node0
+        prev_epoch_block_hash = epoch_block_hash
+        epoch_block_hash, epoch_number = get_epoch_data(scid, self.nodes[0], EPOCH_LENGTH)
+        mark_logs("Height: {}. creation height {}".format(self.nodes[0].getblockcount(), sc_creating_height), self.nodes, DEBUG_MODE)
+        mark_logs("Height: {}. Epoch {}".format(self.nodes[0].getblockcount(), epoch_number), self.nodes, DEBUG_MODE)
+        mark_logs("Create Cert1 with quality 100 and place it in node0", self.nodes, DEBUG_MODE)
+        quality = 100
```
abi87, line 305: I do not understand why this work. We have already confirmed at least one certificate with quality = 100, we should not be able to push another certificate with same quality neither in mempool nor in blockchain

qa/rpc-tests/quality_mempool.py
```diff
@@ -0,0 +1,424 @@
...
+        mark_logs("Check cert is in mempools", self.nodes, DEBUG_MODE)
+        assert_equal(True, cert_1_epoch_0 in self.nodes[0].getrawmempool())
+
+        # Create Cert2 with lower quality, any fee, deps on cert_1 and try to place it in mempool
+        mark_logs("#Checking rejection of cert_2 with same (scId, epoch), lower quality,  any fee, with spending deps on cert_1", self.nodes, DEBUG_MODE)
+        low_quality_proof = mcTest.create_test_proof(
+            "sc1", epoch_number_1, epoch_block_hash_1, prev_epoch_block_hash,
+            quality - 10, constant_1, [pkh_node1], [bwt_amount])
+        try:
+            cert_2_epoch_0 = self.nodes[0].send_certificate(scid_1, epoch_number_1, quality - 10, epoch_block_hash_1, low_quality_proof, amount_cert, CERT_FEE)
```
alsala, line 204: When a dependancy between certs is required, it would be better using the rpc cmd 'createrawcertificate' which allows an explicit settings of the relevant input instead of relying on the src code implementation of the wallet, which implicitly uses previous cert change.

alsala (changes requested):

qa/rpc-tests/quality_voiding.py
```diff
@@ -0,0 +1,330 @@
...
+
+        mark_logs("Checking mempools alignement", self.nodes, DEBUG_MODE)
+        self.sync_all()
+        for i in range(1, NUMB_OF_NODES):
+            assert_equal(sorted(self.nodes[0].getrawmempool()), sorted(self.nodes[i].getrawmempool()))
+
+        mark_logs("Check cert is in mempools", self.nodes, DEBUG_MODE)
+        assert_equal(True, cert_1_epoch_0 in self.nodes[0].getrawmempool())
+
+        # Create Cert2 with lower quality and place it in mempool
```
alsala, line 167: fix comment (higher instead of lower)

qa/rpc-tests/quality_voiding.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # Checking all certificate in mempool
+        self.sync_all()
+        assert_equal(True, cert_3_epoch_0 in self.nodes[0].getrawmempool())
+
+        mined = self.nodes[0].generate(1)[0]
+        self.sync_all()
+
+        # Check balance
+        bwt_address = self.nodes[1].getrawcertificate(cert_3_epoch_0, 1)['vout'][1]['scriptPubKey']['addresses'][0]
```
alsala, line 236: bwt_address must be the same as above at line 195, we can use it or assert they are equal. The same applies in other lines below

qa/pull-tester/rpc-tests.sh
```diff
@@ -93,6 +93,10 @@ testScripts=(
   'sbh_rpc_cmds.py'
   'sc_cert_ceasing.py'
   'sc_getscinfo.py'
+  'quality_mempool.py'
```
alsala, line 96: In order to conform to other test naming, pls rename them adding the 'sc_' prefix

alsala (changes requested):

qa/rpc-tests/quality_voiding.py
```diff
@@ -0,0 +1,330 @@
...
+        assert_equal(res[0]['scid'], scid)
+        mark_logs("Height : {}".format(res[1]['vout'][1]['maturityHeight']), self.nodes, DEBUG_MODE)
+        ud = self.nodes[1].getunconfirmedtxdata(bwt_address)
+        assert_equal(bwt_amount_3, ud['bwtImmatureOutput'])
+        mark_logs("Immature output : {}".format(ud['bwtImmatureOutput']), self.nodes, DEBUG_MODE)
+
+        utxos = self.nodes[1].listunspent()
+        assert_equal(len(utxos), 1)
+        assert_equal(utxos[0]['txid'], cert_3_epoch_0)
+
```
alsala, line 327: We can check that the certificates with lower quality have been properly voided via the command:
    self.nodes[0].gettxout(cert_id, n)
where n is the output number: 0 for the change, never to be voided, and 1 for the bwt, which must be null for all certs but the top quality

alsala (changes requested):

qa/rpc-tests/quality_nodes.py
VolodymyrBabaiev: Quality value resets on each epoch. This certificate doesn't relate to the previous ones.

qa/rpc-tests/quality_blockchain.py
```diff
@@ -0,0 +1,416 @@
...
+        connect_nodes_bi(self.nodes, 1, 2)
+        sync_blocks(self.nodes[1:NUMB_OF_NODES])
+        sync_mempools(self.nodes[1:NUMB_OF_NODES])
+        self.is_network_split = split
+        self.sync_all()
+
+    def run_test(self):
+
+        '''
+        The test creates a sc, send funds to it and then sends a certificate to it,
```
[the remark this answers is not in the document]
VolodymyrBabaiev: Done.

qa/rpc-tests/quality_blockchain.py
```diff
@@ -0,0 +1,416 @@
...
+from test_framework.mc_test.mc_test import *
+import os
+import time
+import pprint
+from decimal import Decimal
+
+DEBUG_MODE = 1
+NUMB_OF_NODES = 3
+EPOCH_LENGTH = 20
+CERT_FEE = Decimal('0.00015')
```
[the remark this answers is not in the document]
VolodymyrBabaiev: Fixed LOW and HIGH values.

qa/rpc-tests/quality_voiding.py
```diff
@@ -0,0 +1,362 @@
...
+        connect_nodes_bi(self.nodes, 0, 1)
+        connect_nodes_bi(self.nodes, 1, 2)
+        sync_blocks(self.nodes[1:NUMB_OF_NODES])
+        sync_mempools(self.nodes[1:NUMB_OF_NODES])
+        self.is_network_split = split
+        self.sync_all()
+
+    def run_test(self):
+
+        '''
```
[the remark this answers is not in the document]
VolodymyrBabaiev: Done.

qa/rpc-tests/quality_voiding.py
VolodymyrBabaiev: Block invalidation and all checking moved to node 1.

qa/rpc-tests/quality_mempool.py
```diff
@@ -0,0 +1,519 @@
...
+            mark_logs("Send certificate failed with reason {}".format(errorString), self.nodes, DEBUG_MODE)
+            assert (False)
+
+        # Checking all certificate in mempool
+        self.sync_all()
+        assert_equal(True, cert_1_epoch_0 in self.nodes[0].getrawmempool())
+        assert_equal(True, cert_2_epoch_0 in self.nodes[0].getrawmempool())
+
+        # Create Cert3 with equal quality but lower fee and try to place it in mempool
+        mark_logs("Checking rejection of cert_3 with same (scId, epoch), equal quality, lower fee, no spending deps on cert_1", self.nodes, DEBUG_MODE)
```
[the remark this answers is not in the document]
VolodymyrBabaiev: Done.

qa/rpc-tests/quality_voiding.py
VolodymyrBabaiev: Done.

qa/rpc-tests/quality_voiding.py
VolodymyrBabaiev: Done.

src/txmempool.cpp
```diff
@@ -1150,3 +1276,112 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+    }
+    for(const auto &c: conflictingCerts) {
+        LogPrint("mempool", "%s():%d - syncing cert %s\n", __func__, __LINE__, c.GetHash().ToString());
+        SyncWithWallets(c, nullptr);
+    }
+
+    return true;
+}
+
+bool CTxMemPool::IsTopQualityCertInMempool(const CScCertificate& cert)
```
i-Alex, line 1328: It is actually the top quality check for the certificates for given `cert.scId`. Let's think about more accurate name.

src/wallet/wallet.cpp
```diff
@@ -2018,7 +2040,17 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
                         sidechain.SafeguardMargin() - nHeight;
 
                     if (CSidechain::State::CEASED == pcoinsTip->isCeasedAtHeight(cert.GetScId(), chainActive.Height()))
-                        areBwtVoided = true;
+                    {
+                        LogPrint("cert", "%s():%d - cert[%s] is CEASED\n", __func__, __LINE__, cert.GetHash().ToString());
+                        areBwtStripped = true;
+                    }
+
+                    if (pcoinsTip->IsBwtStripped(cert.GetHash()) )
```
i-Alex, line 2048: `else if` ?

i-Alex (changes requested):

src/txmempool.cpp
```diff
@@ -1112,24 +1243,19 @@ bool CCoinsViewMemPool::GetSidechain(const uint256& scId, CSidechain& info) cons
...
             for (const auto& fwdAmount : fwdTx.GetVftCcOut())
                 if (scId == fwdAmount.scId)
                     info.mImmatureAmounts[-1] += fwdAmount.nValue;
         }
-
-        if (!mempool.mapSidechains.at(scId).backwardCertificate.IsNull()) {
-            const uint256& certHash = mempool.mapSidechains.at(scId).backwardCertificate;
-            const CScCertificate & cert = mempool.mapCertificate.at(certHash).GetCertificate();
-            info.balance -= cert.GetValueOfBackwardTransfers();
-        }
```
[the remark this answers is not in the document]
i-Alex: We can add to the RPC `getScInfo` method an unconfirmed data section that will contain:
1) unconfirmed certificates info
2) unconfirmed fwd transfers balances
3) in future, BTRs as well.

src/txmempool.cpp
```diff
@@ -1112,24 +1243,19 @@ bool CCoinsViewMemPool::GetSidechain(const uint256& scId, CSidechain& info) cons
...
     } else if (!base->GetSidechain(scId, info))
         return false;
 
-    //decorate sidechain with fwds and bwt in mempool
-    if (mempool.mapSidechains.count(scId)) {
-        for (const auto& fwdHash: mempool.mapSidechains.at(scId).fwdTransfersSet) {
+    //decorate sidechain with fwds and bwt in mempool if any
+    if (mempool.mapSidechains.count(scId) != 0)
+    {
+        for (const auto& fwdHash: mempool.mapSidechains.at(scId).fwdTransfersSet)
```
i-Alex, line 1252: this part can be removed according to the comment below.

src/txmempool.cpp
```diff
@@ -1150,3 +1276,112 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+    const uint256& scId = cert.GetScId();
+    if (mapSidechains.count(scId) != 0  &&
+        !mapSidechains.at(scId).mBackwardCertificates.empty())
+    {
+        return cert.GetHash() == mapSidechains.at(scId).GetTopQualityCert()->second;
+    }
+    return false;
+}
+
+void CTxMemPool::SyncLowQualityCerts(const CScCertificate& cert)
```
i-Alex, line 1339: Where do we use it?

i-Alex (changes requested):

src/miner.cpp
```diff
@@ -86,6 +86,34 @@ class TxPriorityCompare
...
+                        assert("two certs for the same scid can not have the same quality" == 0);
+
+                    return aCert.quality > bCert.quality;
+                }
+            } catch (...) {
+                LogPrintf("%s():%d - ERROR: cast error\n", __func__, __LINE__ );
+                assert("could not cast txbase obj" == 0);
+            }
+        }
+
```
abi87, line 116: While appreciate the succinctness of this change, I think that
1- probably we should not assert into this `operator()`, but write an ad hoc function like 'VerifyCertificateDependencies', similarly to what we have done with `VerifyForwardTransfersDependencies` in this same file
2- 1- we should not assert right away; compare `VerifyForwardTransfersDependencies` where, in some cases we assert only if fDebug flag is set
@i-Alex @alsala what do you think about it?

abi87: Also I'd take advantage to get rid of all the `#if 0` blocks left over from beta :-)

abi87: As agreed offline, I will move checks to a `VerifyCertificateDependencies` and make sure to mimic assert policy used already in  `VerifyForwardTransfersDependencies `

src/wallet/wallet.cpp
abi87: I find `IsBwtStripped` not precise, since, for a given certificate, bwt may be not available because they are normally spent rather than voiding. A future dev may be tricked into thinking that `IsBwtStripped` is able to distinguish voided outputs from spent ones.
I would use instead a method like `HighQualityCertDataFor(block)` to inspect the current block and find all the low quality certs whose bwts must be voided (both in current block and in view)

src/miner.cpp
abi87: Done

src/txmempool.cpp
```diff
@@ -198,147 +232,225 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CCertificateMemPoolEntr
...
+            pCurrentRoot = &mapCertificate.at(toVisit.front()).GetCertificate();
+        } else
+            assert(pCurrentRoot);
+
+        res.insert(toVisit.front());
+        toVisit.pop_front();
+
+        std::set<uint256> directDescendants = mempoolDirectDescendantsOf(*pCurrentRoot);
+        for(const uint256& dep : directDescendants)
+            if (std::find(toVisit.begin(), toVisit.end(),dep) == toVisit.end())
```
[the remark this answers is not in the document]
abi87: Improved termination condition and introduced std::vector instead of std::set to respect topological ordering of txes

src/main.cpp
```diff
@@ -2351,20 +2407,27 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
...
+        const CTxUndo &certUndo = blockUndo.vtxundo[certOffset + i];
+        if (!view.RevertCertOutputs(cert, certUndo, pVoidedCertsMap) ) {
+            LogPrint("sc", "%s():%d - ERROR undoing certificate\n", __func__, __LINE__);
+            return error("DisconnectBlock(): certificate can not be reverted: data inconsistent");
+        }
+
+        // TODO quality
+        // if we have more certs for scid/epoch with different qualities we must handle it somehow
+        // maybe out of the loop on certificates after having collected all of them?
+        // unless consensus rules that certs are ordered by quality
```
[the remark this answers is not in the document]
abi87: Done

src/txmempool.cpp
```diff
@@ -198,147 +232,225 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CCertificateMemPoolEntr
...
+            pCurrentNode = &mapCertificate.at(toVisit.back()).GetCertificate();
+        } else
+            assert(pCurrentNode);
+
+        toVisit.pop_back();
+        res.insert(pCurrentNode->GetHash());
+
+        std::set<uint256> directAncestors = mempoolDirectAncestorsOf(*pCurrentNode);
+        for(const uint256& ancestor : directAncestors) {
+            if (std::find(toVisit.begin(), toVisit.end(), ancestor) == toVisit.end())
```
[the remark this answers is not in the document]
abi87: Done, also introduced std::vector instead of std:.set to preserver topological ordering of txes

src/txmempool.cpp
```diff
@@ -198,147 +232,225 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CCertificateMemPoolEntr
...
+        std::set<uint256> directDescendants = mempoolDirectDescendantsOf(*pCurrentRoot);
+        for(const uint256& dep : directDescendants)
+            if (std::find(toVisit.begin(), toVisit.end(),dep) == toVisit.end())
+                toVisit.push_front(dep);
+    }
+
+    return res;
+}
+
+void CTxMemPool::remove(const CTransactionBase& origTx, std::list<CTransaction>& removedTxs, std::list<CScCertificate>& removedCerts, bool fRecursive)
```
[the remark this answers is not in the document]
abi87: Much appreciated

src/txmempool.cpp
abi87: Done

src/txmempool.h
```diff
@@ -176,9 +187,16 @@ class CTxMemPool
     void check(const CCoinsViewCache *pcoins) const;
     void setSanityCheck(bool _fSanityCheck) { fSanityCheck = _fSanityCheck; }
 
+    std::pair<uint256, CAmount> FindConflictingCert(const uint256& scId, int64_t certQuality);
+    bool RemoveConflictingQualityCert(const uint256& certToRmHash);
```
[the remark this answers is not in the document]
abi87: Renamed to `RemoveCertAndSync`

src/txmempool.h
```diff
@@ -176,9 +187,16 @@ class CTxMemPool
     void check(const CCoinsViewCache *pcoins) const;
     void setSanityCheck(bool _fSanityCheck) { fSanityCheck = _fSanityCheck; }
 
+    std::pair<uint256, CAmount> FindConflictingCert(const uint256& scId, int64_t certQuality);
```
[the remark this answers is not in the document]
abi87: Renamed to `FindCertWithQuality`

src/txmempool.cpp
abi87: Unused, indeed. Removed

src/main.cpp
```diff
@@ -1240,6 +1293,9 @@ bool AcceptCertificateToMemoryPool(CTxMemPool& pool, CValidationState &state, co
             return error("%s(): BUG! PLEASE REPORT THIS! ConnectInputs failed against MANDATORY but not STANDARD flags %s", __func__, certHash.ToString());
         }
 
+        if (!pool.RemoveConflictingQualityCert(conflictingCertData.first))
```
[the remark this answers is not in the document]
abi87: Renamed to `RemoveCertAndSync`; removed assert since wrong;

src/rpc/blockchain.cpp
```diff
@@ -1050,7 +1050,8 @@ UniValue reconsiderblock(const UniValue& params, bool fHelp)
     return NullUniValue;
 }
 
-bool FillScRecordFromInfo(const uint256& scId, const CSidechain& info, CSidechain::State scState, UniValue& sc, bool bOnlyAlive, bool bVerbose)
+bool FillScRecordFromInfo(const uint256& scId, const CSidechain& info, CSidechain::State scState, const CCoinsViewCache& view,
```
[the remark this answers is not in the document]
abi87: Indeed view is not needed anymore since CSidechain state has been extended. Dropped view

src/txmempool.cpp
abi87: Actually not called anymore. Dropped

src/coins.cpp
```diff
@@ -659,6 +661,26 @@ bool CCoinsViewCache::GetSidechain(const uint256 & scId, CSidechain& targetScInf
     return false;
 }
 
+bool CCoinsViewCache::GetTopQualityCert(const uint256& scId, int epochNumber, uint256& hash) const
```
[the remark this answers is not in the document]
abi87: `GetTopQualityCert` dropped in favor of different solution

src/coins.cpp
```diff
@@ -676,6 +698,27 @@ void CCoinsViewCache::GetScIds(std::set<uint256>& scIdsList) const
     return;
 }
 
+bool CCoinsViewCache::IsQualityValid(const CScCertificate& cert, CAmount unused) const
```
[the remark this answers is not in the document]
abi87: Dropped in favor of different solution

src/wallet/wallet.cpp
abi87: Hasty comment above. In given function wallet has been reconstructed from past blocks, while `pCoinsTip `view already refers to `chainActive.Tip()`. As a consequence, we are not fully able to reconstruct reason why cert bwts are voided (normally spent? last cert of ceased sc? low quality cert?).
Currently no best solution that flagging all scenario above with `allBwtsCleared `flag and move on

src/coins.cpp
```diff
@@ -676,6 +698,27 @@ void CCoinsViewCache::GetScIds(std::set<uint256>& scIdsList) const
...
+        {
+            LogPrint("cert", "%s():%d - NOK, cert %s q=%d : a cert q=%d for same sc/epoch is already in blockchain\n",
+                __func__, __LINE__, cert.GetHash().ToString(), cert.quality, info.lastCertificateQuality);
+            return false;
+        }
+    }
+
+    LogPrint("cert", "%s():%d - cert %s q=%d : OK, no better quality certs for same sc/epoch are in blockchain\n",
+        __func__, __LINE__, cert.GetHash().ToString(), cert.quality);
+    return true;
```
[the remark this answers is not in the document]
abi87: Superseeded by other changes

src/coins.cpp
```diff
@@ -676,6 +698,27 @@ void CCoinsViewCache::GetScIds(std::set<uint256>& scIdsList) const
...
 }
 
+bool CCoinsViewCache::IsQualityValid(const CScCertificate& cert, CAmount unused) const
+{
+    // check in blockchain if a better cert is already there for this epoch
+    CSidechain info;
+    if (GetSidechain(cert.GetScId(), info))
+    {
+        if (info.lastEpochReferencedByCertificate == cert.epochNumber &&
+            info.lastCertificateQuality >= cert.quality)
```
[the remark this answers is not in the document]
abi87: Superseeded by other changes

src/txmempool.cpp
abi87: Superseeded by other changes

src/coins.cpp
```diff
@@ -1290,6 +1521,27 @@ bool CCoinsViewCache::CancelSidechainEvent(const CScCertificate& cert)
         return false;
     }
 
+    //LogPrint("sc", "%s():%d -------------- info -------------- \n", __func__, __LINE__);
+    //Dump_info();
+
+    if (restoredScInfo.topCommittedCertReferencedEpoch == cert.epochNumber)
```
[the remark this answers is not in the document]
abi87: Done

src/main.cpp
```diff
@@ -2351,20 +2407,27 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
             }
   
             // remove outputs
-            LogPrint("cert", "%s():%d - clearing outs of tx[%s]\n", __func__, __LINE__, hash.ToString());
+            LogPrint("cert", "%s():%d - clearing outs of cert[%s]\n", __func__, __LINE__, hash.ToString());
             outs->Clear();
         }
 
+        // this must be called before cancelling sc events since certificate quality is handled here
```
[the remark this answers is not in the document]
abi87: Done

src/txmempool.cpp
```diff
@@ -1150,3 +1227,169 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+    if (!base->IsQualityValid(cert, certFee))
+    {
+        LogPrint("cert", "%s:%s():%d - cert %s NOK\n", __FILE__, __func__, __LINE__, cert.GetHash().ToString());
+        return false;
+    }
+
+    const uint256& scId = cert.GetScId();
+
+    if ((mempool.mapSidechains.count(scId) != 0) &&
+        (!mempool.mapSidechains.at(scId).mBackwardCertificates.size() == 0))
```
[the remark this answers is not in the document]
abi87: Superseeded by other code changes

src/coins.cpp
```diff
@@ -1027,31 +1077,165 @@ bool CCoinsViewCache::UpdateScInfo(const CScCertificate& cert, CTxUndo& certUndo
...
+        {
+            coins->vout.resize(coins->nFirstBwtPos + idx+1);
+        }
+        coins->vout.at(coins->nFirstBwtPos + idx) = outsToRestore.at(idx).txout;
+    }
+ 
+    return fClean;
+}
+
+bool CCoinsViewCache::RevertCertOutputs(const CScCertificate& cert, const CSidechainUndoData& sidechainUndo)
```
albertog78, line 1235: RestoreSCInfo(certToRevert, scUndoData)

src/coins.cpp
```diff
@@ -1063,13 +1247,40 @@ bool CCoinsViewCache::RevertCertOutputs(const CScCertificate& cert, const CTxUnd
...
 
-    LogPrint("cert", "%s():%d - amount restored to scView (amount=%s, resulting bal=%s) %s\n",
-        __func__, __LINE__, FormatMoney(totalAmount), FormatMoney(scIt->second.scInfo.balance), scId.ToString());
+    // restore only if this is the top quality cert for this epoch
+    LogPrint("cert", "%s():%d - cert %s, last cert %s, undo cert %s\n", __func__, __LINE__,
+        cert.GetHash().ToString(), scIt->second.scInfo.topCommittedCertHash.ToString(), sidechainUndo.prevTopCommittedCertHash.ToString());
+    LogPrint("cert", "%s():%d - cert epoch %d, last epoch %d, undo epoch %d\n", __func__, __LINE__,
+        cert.epochNumber, scIt->second.scInfo.topCommittedCertReferencedEpoch, sidechainUndo.prevTopCommittedCertReferencedEpoch);
+
+    if (cert.GetHash() == scIt->second.scInfo.topCommittedCertHash)
```
albertog78, line 1257: modify this into an assert (this method must be call only for block top quality certs)

src/main.cpp
```diff
@@ -1980,7 +2090,7 @@ void UpdateCoins(const CTransaction& tx, CCoinsViewCache &inputs, CTxUndo &txund
     inputs.ModifyCoins(tx.GetHash())->From(tx, nHeight);
 }
 
-void UpdateCoins(const CScCertificate& cert, CCoinsViewCache &inputs, CTxUndo &txundo, int nHeight)
+void UpdateCoins(const CScCertificate& cert, CCoinsViewCache &inputs, CTxUndo &txundo, int nHeight, bool fVoidBwts)
```
albertog78, line 2093: bool isBlockTopQualityCert

src/gtest/tx_creation_utils.cpp
```diff
@@ -164,6 +166,19 @@ CScCertificate txCreationUtils::createCertificate(const uint256 & scId, int epoc
     return res;
 }
```
albertog78, line 168: reached this point

albertog78 (changes requested):

src/txmempool.cpp
```diff
@@ -198,147 +233,228 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CCertificateMemPoolEntr
     return true;
 }
 
-void CTxMemPool::remove(const CTransactionBase& origTx, std::list<CTransaction>& removedTxs, std::list<CScCertificate>& removedCerts, bool fRecursive)
+std::vector<uint256> CTxMemPool::mempoolDirectAncestorsOf(const CTransactionBase& root) const
```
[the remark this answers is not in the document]
abi87: Done

src/coins.cpp
```diff
@@ -1027,31 +1077,165 @@ bool CCoinsViewCache::UpdateScInfo(const CScCertificate& cert, CTxUndo& certUndo
     }
 
     CSidechainsMap::iterator scIt = ModifySidechain(scId);
-    if (scIt->second.scInfo.balance < totalAmount)
+
+    blockUndo.scUndoDatabyScId[scId].prevTopCommittedCertReferencedEpoch = scIt->second.scInfo.topCommittedCertReferencedEpoch;
```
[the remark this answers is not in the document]
abi87: Done

src/main.cpp
```diff
@@ -2334,6 +2459,13 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
             int currentEpoch = sidechain.EpochFor(pindex->nHeight);
             int bwtMaturityHeight = sidechain.StartHeightForEpoch(currentEpoch+1) + sidechain.SafeguardMargin();
             CCoins outsBlock(cert, pindex->nHeight, bwtMaturityHeight);
+            if (highQualityCertData.count(cert.GetHash()) == 0) //drop bwts of low q certs
```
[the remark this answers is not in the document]
abi87: Done

src/main.cpp
```diff
@@ -2300,21 +2428,18 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
...
+    if (!view.RevertSidechainEvents(blockUndo, pindex->nHeight, pVoidedCertsMap))
     {
         LogPrint("cert", "%s():%d - SIDECHAIN-EVENT: failed reverting scheduled event\n", __func__, __LINE__);
         return error("DisconnectBlock(): cannot revert sidechains scheduled events");
     }
 
-
     // not including coinbase
     const int certOffset = block.vtx.size() - 1;
+    std::map<uint256,uint256> highQualityCertData = HighQualityCertData(block, blockUndo);
```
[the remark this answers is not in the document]
abi87: Done

src/undo.h
```diff
@@ -136,64 +108,163 @@ class CTxUndo
...
-struct ScUndoData
+struct CSidechainUndoData
 {
+    uint32_t sidechainUndoDataVersion;
+    enum AvailableSections : uint8_t
+    {
+        UNDEFINED               = 0,
+        SIDECHAIN_STATE         = 1,
+        MATURED_AMOUNTS         = 2,
+        LOW_QUALITY_CERT_DATA   = 4,
```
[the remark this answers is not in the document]
abi87: Done

src/coins.cpp
abi87: Done

src/coins.cpp
abi87: Done

src/main.cpp
abi87: Done

src/txmempool.cpp
```diff
@@ -854,7 +985,7 @@ void CTxMemPool::check(const CCoinsViewCache *pcoins) const
         } else {
             assert(::ContextualCheckInputs(entry->GetCertificate(), state, mempoolDuplicateCert, false, chainActive, 0, false, Params().GetConsensus(), NULL));
             CTxUndo dummyUndo;
-            UpdateCoins(entry->GetCertificate(), mempoolDuplicateCert, dummyUndo, 1000000);
+            UpdateCoins(entry->GetCertificate(), mempoolDuplicateCert, dummyUndo, 1000000, /*fVoidBwts*/false);
```
albertog78, line 988: true

src/txmempool.cpp
```diff
@@ -1085,7 +1216,7 @@ bool CCoinsViewMemPool::GetCoins(const uint256 &txid, CCoins &coins) const {
     CScCertificate cert;
     if (mempool.lookup(txid, cert)) {
         LogPrint("mempool", "%s():%d - making coins for cert [%s]\n", __func__, __LINE__, txid.ToString() );
-        coins = CCoins(cert, MEMPOOL_HEIGHT, MEMPOOL_HEIGHT);
+        coins = CCoins(cert, MEMPOOL_HEIGHT, MEMPOOL_HEIGHT, /*isBlockTopQualityCert*/true);
```
albertog78, line 1219: check if it is really top quality

src/wallet/rpcwallet.cpp
```diff
@@ -4845,7 +4845,10 @@ UniValue send_certificate(const UniValue& params, bool fHelp)
     scId.SetHex(scIdString);
 
     // sanity check of the side chain ID
-    CCoinsViewCache scView(pcoinsTip);
+    CCoinsView dummy;
```
albertog78, line 4848: reached this point

albertog78 (changes requested):

src/main.cpp
```diff
@@ -2333,7 +2443,9 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
             assert(view.GetSidechain(cert.GetScId(), sidechain));
             int currentEpoch = sidechain.EpochFor(pindex->nHeight);
             int bwtMaturityHeight = sidechain.StartHeightForEpoch(currentEpoch+1) + sidechain.SafeguardMargin();
-            CCoins outsBlock(cert, pindex->nHeight, bwtMaturityHeight);
+            CCoins outsBlock(cert, pindex->nHeight, bwtMaturityHeight, isBlockTopQualityCert);
+            outsBlock.ClearUnspendable();
```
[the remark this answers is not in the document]
abi87: Done

src/txmempool.cpp
abi87: Done

src/txmempool.cpp
abi87: Done

albertog78 (approved):

PR #61, opened 2020-12-01
Quality check refactoring

abi87:
Refactored minimal quality check and conflicts in mempool.
Importantly, it forces quality to be always increasing across blocks, including blocks belonging to different epochs


abi87:
As discussed offline, quality should not be ever increasing but increasing within single epochs.
Closing this one since not compliant with specifications

PR #64, opened 2020-12-09
Skip Sidechain state update for low quality certificates

abi87:
with UTs


src/main.cpp
```diff
@@ -2390,6 +2408,13 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
             int currentEpoch = sidechain.EpochFor(pindex->nHeight);
             int bwtMaturityHeight = sidechain.StartHeightForEpoch(currentEpoch+1) + sidechain.SafeguardMargin();
             CCoins outsBlock(cert, pindex->nHeight, bwtMaturityHeight);
+            if (highQualityCertData.count(cert.GetHash()) == 0) //drop bwts of low q certs
```
i-Alex, line 2411: why do we need to drop bwts of low quality certificates on disconnect block? Were they added to coins view on connect block?

abi87: In this version of quality implementation, bwts of low quality certificates are dropped right at coin creation and no block undo is created for them (they won't ever be restored, not even on block disconnect).
In this code section, lines 2402-2437 a coherence check is being carried out: essentially we are verifying that the coins that would be created out of the block being disconnected match the ones already in view (which were created by the very same block)
That is why I need to void bwts of low quality certs, similarly to what I have done in block connect.

i-Alex (approved):

alsala (approved):

PR #66, opened 2020-12-10
Minor changes to block formation algo + UTs

abi87:
Just introduced `VerifyCertificatesDependencies`

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -52,82 +52,44 @@ using namespace zen;
 // BitcoinMiner
 //
 
-//
-// Unconfirmed transactions in the memory pool often depend on other
-// transactions in the memory pool. When we select transactions from the
-// pool, we select by highest priority or fee rate, so we might consider
-// transactions that depend on transactions that aren't yet in the block.
-// The COrphan class keeps track of these 'temporary orphans' while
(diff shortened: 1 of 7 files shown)


src/miner.cpp
```diff
@@ -52,95 +52,97 @@ using namespace zen;
...
-            try {
-                const CScCertificate& aCert = dynamic_cast<const CScCertificate&>(*a.get<2>());
-                const CScCertificate& bCert = dynamic_cast<const CScCertificate&>(*b.get<2>());
- 
-                if (aCert.GetScId() == bCert.GetScId() )
-                {
-                    LogPrint("sc", "%s():%d - qa=%d - qb=%d\n", __func__, __LINE__, aCert.quality, bCert.quality );
-
-                    if (aCert.epochNumber != bCert.epochNumber)
-                        assert("two certs for the same scid can not have different epochs" == 0);
```
alsala, line 104: The check on epoch is not performed in the modified version; we can probably live without it since that should already have been performed in mempool

alsala (approved):

src/miner.cpp
abi87: I believe that check is not essential to the right ordering of certs in block. I would leave it to the mempool as you mentioned

PR #70, opened 2020-12-15
Fixed ceasing sc height limit

alsala:


src/sc/sidechain.cpp
```diff
@@ -48,9 +48,10 @@ std::string CSidechain::stateToString(State s)
 {
     switch(s)
     {
-        case State::ALIVE:  return "ALIVE";          break;
-        case State::CEASED: return "CEASED";         break;
-        default:            return "NOT_APPLICABLE"; break;
+        case State::MEMPOOL:  return "MEMPOOL";        break;
```
abi87, line 51: how about "Unconfirmed" rather than "mempool" as name for the state?

abi87 (changes requested):

src/sc/sidechain.cpp
alsala: ok, done

i-Alex:
Absorbed by the https://github.com/HorizenOfficial/zend_oo/pull/77

PR #73, opened 2020-12-22
Cert quality wallet fix for rescanning

abi87:
Following final code review, extended wallet with minimal sidechain state in order to be able to duly void low quality certificates upon rescanning

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2491,7 +2491,7 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
                 }
             }
 
-            //Refresh previous certificate in wallet, whether it has been just restored or it from previous epoch
+            //Refresh previous certificate in wallet, whether it has been just restored or it is from previous epoch
             if (pCertsStateInfo!= nullptr)
             {
                 pCertsStateInfo->push_back(CScCertificateStatusUpdateInfo(cert.GetScId(),
@@ -2907,7 +2907,6 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     for (unsigned int certIdx = 0; certIdx < block.vcert.size(); certIdx++) // Processing certificates loop
     {
         const CScCertificate &cert = block.vcert[certIdx];
-        LogPrint("cert", "%s():%d - Connecting certificate [%s]\n", __func__, __LINE__, cert.GetHash().ToString());
         nSigOps += GetLegacySigOpCount(cert);
         if (nSigOps > MAX_BLOCK_SIGOPS)
             return state.DoS(100, error("ConnectBlock(): too many sigops"),
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3293,11 +3293,16 @@ bool static DisconnectTip(CValidationState &state) {
     // 0-confirmed or conflicted:
     for(const CTransaction &tx: block.vtx) {
         SyncWithWallets(tx, nullptr);
+        for(unsigned int pos = 0; pos < tx.GetVscCcOut().size(); ++pos)
+            SyncWithWallets(tx.GetScIdFromScCcOut(pos), CMinimalSidechain());
     }
 
     for(const CScCertificate &cert: block.vcert) {
         LogPrint("cert", "%s():%d - sync with wallet from block to unconfirmed cert[%s]\n", __func__, __LINE__, cert.GetHash().ToString());
+        CSidechain sidechain;
+        assert(pcoinsTip->GetSidechain(cert.GetScId(), sidechain));
         SyncWithWallets(cert, nullptr);
+        SyncWithWallets(cert.GetScId(), CMinimalSidechain(sidechain));
     }
 
     for(const auto& x : voidedCertsMap)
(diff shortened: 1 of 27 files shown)


src/wallet/wallet.cpp
```diff
@@ -1330,7 +1330,36 @@ void CWallet::SyncVoidedCert(const uint256& certHash, bool bwtAreStripped)
...
+
+bool CWallet::ReadSidechain(const uint256& scId, CMinimalSidechain& sidechain)
+{
+    bool res = false;
+    if (mapSidechains.count(scId) != 0) {
+        sidechain = mapSidechains.at(scId);
+        res = true;
+    } else {
+        CWalletDB walletdb(strWalletFile, "r+", false);
+        res = walletdb.ReadSidechain(scId, sidechain);
```
alsala, line 1359: if res is false can we rely on retrieved value?

src/wallet/wallet.cpp
```diff
@@ -2021,43 +2030,43 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
...
-                    // which we are not able to reconstruct here.
-                    // Currently we just void all the bwts here
-                    if (pcoinsTip->IsBwtStripped(cert.GetHash()) )
-                    {
-                        LogPrint("cert", "%s():%d - cert[%s] has bwt stripped\n",
-                            __func__, __LINE__, cert.GetHash().ToString());
-                        areBwtStripped = true;
-                    }
-                }
+                if (AddToWalletIfInvolvingMe(tx, &block, -1, fUpdate))
```
alsala, line 2039: Consider syncing only if AddToWalletIfInvolvingMe() returns true

src/sc/sidechain.h
```diff
@@ -155,4 +155,47 @@ namespace Sidechain {
     bool hasScCreationOutput(const CTransaction& tx, const uint256& scId);
 }; // end of namespace
 
+
+struct CMinimalSidechain //Sidechain with minimal info, to be included in wallet
```
alsala, line 159: Consider renaming to CWalletSidechain

alsala (changes requested):

src/sc/sidechain.h
abi87: I see the value of the suggestion. Let's discuss during final code review

src/wallet/wallet.cpp
abi87: Unlike SyncVoidingCert, which requires cert to be added to wallet map before being voided, syncSidechain is independent.

src/wallet/wallet.cpp
abi87: No, we cannot. Fixed

src/sc/sidechain.h
```diff
@@ -155,4 +155,47 @@ namespace Sidechain {
...
+    CMinimalSidechain(int32_t certEpoch, uint256 certHash):
+        prevBlockTopQualityCertReferencedEpoch(certEpoch),
+        prevBlockTopQualityCertHash(certHash) {};
+
+    bool IsNull() const {
+        return (prevBlockTopQualityCertReferencedEpoch == CScCertificate::EPOCH_NOT_INITIALIZED) &&
+               (prevBlockTopQualityCertHash.IsNull());
+    }
+
+    size_t GetSerializeSize(int nType, int nVersion) const {
```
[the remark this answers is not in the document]
abi87: Agree, done

src/main.cpp
```diff
@@ -3387,7 +3393,15 @@ bool static ConnectTip(CValidationState &state, CBlockIndex *pindexNew, CBlock *
     for(const CTransaction &tx: pblock->vtx) {
         LogPrint("cert", "%s():%d - sync with wallet tx[%s]\n", __func__, __LINE__, tx.GetHash().ToString());
         SyncWithWallets(tx, pblock);
+
```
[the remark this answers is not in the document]
abi87: I see your concern. Let's discuss during final code review

src/main.cpp
```diff
@@ -3293,15 +3293,21 @@ bool static DisconnectTip(CValidationState &state) {
     // 0-confirmed or conflicted:
     for(const CTransaction &tx: block.vtx) {
         SyncWithWallets(tx, nullptr);
+        for(unsigned int pos = 0; pos < tx.GetVscCcOut().size(); ++pos)
```
albertog78, line 3296: Instead of introducing a new signal, let's use the existing SyncVoidedCert (updating the name to SyncCertStatusUpdate).
To be able to update in the wallet the sc with the current certId and epoch we should also signal the following case:
* last existing top quality cert of an epoch is disconnected (this cert is already communicated as voided)
* the sidechain has a previous epoch and a cert for it
We need to communicate a sync event also for this prev epoch certificate with the status BWT_ON (in this way wallet will be able to update last sidechain cert id and epoch with this one).

Modify the voidedCertsMap in a set of objects (CCertificateStatusUpdateInfo) each containing:
* certHash
* epoch
* sidechainId
* enum {BWT_ON, BWT_OFF}

Drop notification of sc creation

albertog78 (changes requested):

albertog78 (changes requested):

albertog78 (approved):

PR #77, opened 2021-01-05
Ceased sidechain withdrawal

i-Alex:


abi87 (changes requested):

src/txmempool.cpp
```diff
@@ -180,12 +180,18 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CTxMemPoolEntry &entry,
...
-        BOOST_FOREACH(const uint256 &nf, joinsplit.nullifiers) {
+    for(const JSDescription &joinsplit: tx.GetVjoinsplit()) {
+        for(const uint256 &nf: joinsplit.nullifiers) {
             mapNullifiers[nf] = &tx;
         }
     }
 
+    for(const CTxCeasedSidechainWithdrawalInput& csw: tx.GetVcswCcIn()) {
+        if (mapSidechains.count(csw.scId) == 0)
+            LogPrint("mempool", "%s():%d - adding [%s] in mapSidechain\n", __func__, __LINE__, csw.scId.ToString() );
```
abi87, line 191: We have a better trace in mbtr_introduction branch. Let's remember to align the traces.

src/txmempool.cpp
```diff
@@ -688,6 +704,23 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
...
+            if(cswNullifierPos != mapSidechains.at(csw.scId).cswNullifiers.end()) {
+                const uint256& txHash = cswNullifierPos->second;
+                const auto& it = mapTx.find(txHash);
+                if (it != mapTx.end()) {
+                    const CTransaction &txConflict = it->second.GetTx();
+                    if (txConflict != tx)
+                    {
+                        remove(txConflict, removedTxs, removedCerts, true);
+                    }
+                }
```
abi87, line 720: I wonder if the `if (it != mapTx.end()) {` shouldn't be an assert. I mean you should be guaranteed that, if you find a nullifier in mempool, the tx containing it is in mempool as well right?
If so, either assert, or use directly 'mapTx.at(txHash)'

src/txmempool.cpp
```diff
@@ -861,11 +897,21 @@ void CTxMemPool::check(const CCoinsViewCache *pcoins) const
             const auto& fwdPos = mapSidechains.at(fwd.scId).fwdTransfersSet.find(tx.GetHash());
             assert(fwdPos != mapSidechains.at(fwd.scId).fwdTransfersSet.end());
 
-            //there must be no dangling fwds, i.e. sc creation is either in mempool or in blockchain
+            //there must be no dangling fwds, i.e. sc creation is either in mempool or in blockchain (and not ceased)
             if (!mapSidechains.at(fwd.scId).scCreationTxHash.IsNull())
                 assert(mapTx.count(mapSidechains.at(fwd.scId).scCreationTxHash));
             else
-                assert(pcoins->HaveSidechain(fwd.scId));
+                assert(pcoins->GetSidechainState(fwd.scId) == CSidechain::State::ALIVE);
```
abi87, line 904: This constraint in fact was already true (and unchecked), right?

src/gtest/test_mempool.cpp
```diff
@@ -134,13 +134,27 @@ TEST(Mempool, TxInputLimit) {
...
     // Clear the limit
     mapArgs.erase("-mempooltxinputlimit");
 
     // Check it no longer fails due to exceeding the limit
     CValidationState state4;
     EXPECT_FALSE(AcceptTxToMemoryPool(pool, state4, tx3, false, &missingInputs));
     EXPECT_EQ(state4.GetRejectReason(), "bad-txns-version-too-low");
+    CValidationState state4csw;
+    EXPECT_FALSE(AcceptTxToMemoryPool(pool, state4csw, txWithCsw, false, &missingInputs));
+    EXPECT_EQ(state4.GetRejectReason(), "bad-txns-version-too-low");
```
abi87, line 157: I would be great to provide a positive case, where a tx with csw is accepted to mempool. This is especially important if we cannot check the reason for rejection (how are you sure it's not rejected for an unforeseen reason?)

abi87 (changes requested):

src/primitives/transaction.cpp
```diff
@@ -504,6 +550,27 @@ bool CTransactionBase::CheckInputsDuplication(CValidationState &state) const
...
+    bool res = CTransactionBase::CheckInputsDuplication(state);
+    if(!res)
+        return res;
+
+    // Check for duplicate ceased sidechain withdrawal inputs
+    // CSW nullifiers expected to be unique
+    std::set<libzendoomc::ScFieldElement> vNullifiers;
+    for(const CTxCeasedSidechainWithdrawalInput cswIn: GetVcswCcIn())
+    {
+        if(!vNullifiers.insert(cswIn.nullifier).second)
```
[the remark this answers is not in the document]
i-Alex: ok. I will turn back to the style of  `CheckInputsDuplication`

src/primitives/transaction.cpp
```diff
@@ -979,7 +1089,7 @@ bool CTransaction::ContextualCheckInputs(CValidationState &state, const CCoinsVi
           const CChain& chain, unsigned int flags, bool cacheStore, const Consensus::Params& consensusParams,
           std::vector<CScriptCheck> *pvChecks) const
 {
-    return ::ContextualCheckInputs(*this, state, view, fScriptChecks, chain, flags, cacheStore, consensusParams, pvChecks);
+    return ::ContextualCheckTxInputs(*this, state, view, fScriptChecks, chain, flags, cacheStore, consensusParams, pvChecks);
```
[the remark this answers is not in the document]
i-Alex: Let's discuss. The problem, as usual, is in the Miner logic, where we have the list of CTransactionBase.

src/sc/sidechain.cpp
```diff
@@ -61,7 +62,7 @@ size_t CSidechain::DynamicMemoryUsage() const {
 size_t CSidechainEvents::DynamicMemoryUsage() const {
     return memusage::DynamicUsage(maturingScs) + memusage::DynamicUsage(ceasingScs);
 }
-
+// TODO: add CSW verification rules
```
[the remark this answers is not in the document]
i-Alex: sure

src/primitives/transaction.cpp
```diff
@@ -749,6 +835,36 @@ std::string CTransaction::ToString() const
...
+    if (limit > 0) {
+        size_t n = GetVin().size();
+        if (n > limit) {
+            LogPrint("mempool", "Dropping txid %s : too many transparent inputs %zu > limit %zu\n",
+                    GetHash().ToString(), n, limit );
+            return false;
+        }
+    }
+    return true;
+}
```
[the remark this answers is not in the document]
i-Alex: agree

src/script/interpreter.cpp
```diff
@@ -1075,10 +1098,16 @@ class CTransactionSignatureSerializer {
...
 
             // Serialize vin
-            unsigned int nInputs = fAnyoneCanPay ? 1 : txTo.GetVin().size();
+            // In case of SIGHASH_ANYONECANPAY:
+            // * if Tx has Sc support version and nIn belongs to the CSW inputs - skip inputs serialization
+            // * otherwise only the input being signed is serialized
+            // Otherwise - serialize all inputs
+            unsigned int nInputs = fAnyoneCanPay ?
+                        (txTo.IsScVersion() && nIn >= txTo.GetVin().size() ? 0 : 1) :
+                        txTo.GetVin().size();
```
abi87, line 1107: This is a bit hard for me to read. Aren't you say essentially
`if ( ! (txTo.IsScVersion() && nIn >= txTo.GetVin().size()) {/* Do as before*/}
     /*else DO NOTHING*/`?
If so, I would rewrite this as above, for better readability

src/txmempool.cpp
i-Alex: I think that assert is not fully correct. See the logic above with `vin`.
For example, current Tx spends CSWs A and B and there is a conflict Tx' with the same CSWs A and B.
For CSW A we will find and remove Tx', and for B - will skip (because Tx' was removed before).

src/script/interpreter.cpp
```diff
@@ -1087,9 +1116,23 @@ class CTransactionSignatureSerializer {
...
+                            txTo.GetVcswCcIn().size();
+                ::WriteCompactSize(s, nCswInputs);
+                for (unsigned int nCswInput = 0; nCswInput < nCswInputs; nCswInput++)
+                    SerializeCswInput(s, nCswInput, txTo, nType, nVersion);
+
                 // Serialize vccouts
                 unsigned int nCcOutputs = 0;
  
+                // Q: In case of fHashSingle we also serialize all CC outputs. Is it correct?
+                // Seems or we need to ingore all CC outputs, or consider them in a unique joined index of(vouts + cc_outs)
```
abi87, line 1135: I would consider CCOuts as a unique index of `vouts + cc_outs`. The reference for my (currently limited) understanding of signatures is [this](https://raghavsood.com/blog/2018/06/10/bitcoin-signature-types-sighash). In here it is stated that `SIGHASH_SINGLE` should cover all inputs and just one output, expressing the intention "I want to spend my coins only if some other parties are making a transfer too". This interpretation extends naturally to ccOutput as well and suggest the unique index solution

src/script/interpreter.cpp
```diff
@@ -1087,9 +1116,23 @@ class CTransactionSignatureSerializer {
...
+                // In case of SIGHASH_ANYONECANPAY:
+                // * if Tx has Sc support version and nIn belongs to the CSW inputs - only the CSW input being signed is serialized
+                // * otherwise skip CSW inputs
+                // Otherwise - serialize all CSW inputs
+                unsigned int nCswInputs = fAnyoneCanPay ?
+                            (nIn < txTo.GetVin().size() ? 0 : 1) :
+                            txTo.GetVcswCcIn().size();
+                ::WriteCompactSize(s, nCswInputs);
+                for (unsigned int nCswInput = 0; nCswInput < nCswInputs; nCswInput++)
+                    SerializeCswInput(s, nCswInput, txTo, nType, nVersion);
```
abi87, line 1129: Why you have chosen to serialize cws inputs after outputs?
Why don't we process all inputs (ordinary and cws) first and all outputs after (ordinary and cc ones)?

src/coins.cpp
```diff
@@ -1094,6 +1172,7 @@ bool CCoinsViewCache::UpdateScInfo(const CScCertificate& cert, CBlockUndo& block
     blockUndo.scUndoDatabyScId[scId].prevTopCommittedCertQuality         = scIt->second.scInfo.prevBlockTopQualityCertQuality;
     blockUndo.scUndoDatabyScId[scId].prevTopCommittedCertBwtAmount       = scIt->second.scInfo.prevBlockTopQualityCertBwtAmount;
     blockUndo.scUndoDatabyScId[scId].contentBitMask |= CSidechainUndoData::AvailableSections::SIDECHAIN_STATE;
+    blockUndo.scUndoDatabyScId[scId].prevCertDataHash = calculateCertDataHash(cert);
```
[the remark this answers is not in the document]
i-Alex: Here, we should update blockUndo structure with the previous `scInfo` data. So:
`blockUndo.scUndoDatabyScId[scId].prevCertDataHash = scIt->second.scInfo.prevCertDataHash;`

The update for `scInfo` must be done after.

src/coins.cpp
```diff
@@ -1923,7 +2023,7 @@ double CCoinsViewCache::GetPriority(const CTransactionBase &tx, int nHeight) con
             dResult += coins->vout[txin.prevout.n].nValue * (nHeight-coins->nHeight);
         }
     }
-
+    // Q: do we need to consider CSW inputs? set the coin height = 1?
```
i-Alex, line 2026: If we consider the CSW coin age equals to 0 -> CSW inputs decrease the priority of the Tx.
@abi87 what do you think about this?

src/coins.h
```diff
@@ -567,6 +567,8 @@ class CCoinsViewCache : public CCoinsViewBacked
     bool UpdateScInfo(const CTransaction& tx, const CBlock&, int nHeight);
     bool RevertTxOutputs(const CTransaction& tx, int nHeight);
 
+    //CSW INPURS RELATED PUBLIC MEMBERS
```
i-Alex, line 570: -> INPUTS

src/coins.cpp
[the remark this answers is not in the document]
i-Alex: I guess, that we need to have the opposite operation in `CCoinsViewCache::RestoreScInfo` method.

src/main.cpp
```diff
@@ -871,17 +871,29 @@ bool AreInputsStandard(const CTransactionBase& txBase, const CCoinsViewCache& ma
...
     }
-    BOOST_FOREACH(const CTxOut& txout, tx.GetVout())
+    for(const CTxOut& txout: txBase.GetVout())
     {
         nSigOps += txout.scriptPubKey.GetSigOpCount(false);
     }
+
+    if(!txBase.IsCertificate())
+    {
+        const CTransaction& tx = dynamic_cast<const CTransaction&>(txBase);
```
i-Alex, line 888: check formally that `tx` is not null

src/primitives/certificate.cpp
```diff
@@ -232,6 +237,34 @@ std::shared_ptr<const CTransactionBase> CScCertificate::MakeShared() const
...
+{
+    if (nIn >= GetVin().size() )
+        return ::error("%s:%d can not verify Signature: nIn too large for vin size %d",
+                                       GetHash().ToString(), nIn, GetVin().size());
+
+    const CScript &scriptSig = GetVin()[nIn].scriptSig;
+
+    if (!::VerifyScript(scriptSig, scriptPubKey, nFlags,
+                      //CachingTransactionSignatureChecker(this, nIn, chain, cacheStore),
+                      *MakeSignatureChecker(nIn, chain, cacheStore),
```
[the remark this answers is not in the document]
i-Alex: I tried to do it before, but had an error with bitcoin-tx, maybe I missed something. I will check and try again.

src/primitives/certificate.cpp
```diff
@@ -262,6 +295,31 @@ CAmount CScCertificate::GetValueOfChange() const
     return nValueOut;
 }
 
+libzendoomc::ScFieldElement calculateCertDataHash(const CScCertificate& cert) {
```
i-Alex, line 298: We can put these 2 methods to the `proofverifier` file or create a new one in `sc` dir.

src/primitives/certificate.cpp
```diff
@@ -262,6 +295,31 @@ CAmount CScCertificate::GetValueOfChange() const
...
+    // TODO Replace with hash of merkel tree supplied by SC.
+    libzendoomc::ScFieldElement certDataHash;
+    certDataHash.SetHex(cert.GetHash().GetHex());
+    return certDataHash;
+}
+
+libzendoomc::ScFieldElement calculateCumulativeCertDataHash(const libzendoomc::ScFieldElement& prevCumulativeHash, const libzendoomc::ScFieldElement& prevCertHash) {
+    auto prevCumulativeHashInput = zendoo_deserialize_field(prevCumulativeHash.begin());
```
i-Alex, line 306: same as in `CScWCertProofVerification::verifyScCert`we need to check `prevCumulativeHashInput` for `nullptr`

src/primitives/transaction.cpp
```diff
@@ -528,27 +595,28 @@ bool CTransaction::CheckInputsInteraction(CValidationState &state) const
...
 
 CTransaction& CTransaction::operator=(const CTransaction &tx) {
     CTransactionBase::operator=(tx);
-    *const_cast<std::vector<JSDescription>*>(&vjoinsplit)        = tx.vjoinsplit;
-    *const_cast<uint32_t*>(&nLockTime)                           = tx.nLockTime;
-    *const_cast<std::vector<CTxScCreationOut>*>(&vsc_ccout)      = tx.vsc_ccout;
-    *const_cast<std::vector<CTxForwardTransferOut>*>(&vft_ccout) = tx.vft_ccout;
-    *const_cast<uint256*>(&joinSplitPubKey)                      = tx.joinSplitPubKey;
-    *const_cast<joinsplit_sig_t*>(&joinSplitSig)                 = tx.joinSplitSig;
+    *const_cast<std::vector<JSDescription>*>(&vjoinsplit)                    = tx.vjoinsplit;
```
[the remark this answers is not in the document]
i-Alex: it is here. 2 lines below

src/primitives/transaction.cpp
```diff
@@ -749,6 +833,36 @@ std::string CTransaction::ToString() const
...
+}
+
+bool CTransaction::CheckInputsLimit() const {
+    // Node operator can choose to reject tx by number of transparent inputs and csw inputs
+    static_assert(std::numeric_limits<size_t>::max() >= std::numeric_limits<int64_t>::max(), "size_t too small");
+    size_t limit = (size_t) GetArg("-mempooltxinputlimit", 0);
+    if (limit > 0) {
+        size_t n = GetVin().size() + GetVcswCcIn().size();
+        if (n > limit) {
+            LogPrint("mempool", "Dropping txid %s : too many transparent inputs %zu > limit %zu\n",
```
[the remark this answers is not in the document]
i-Alex: ok

src/primitives/transaction.h
```diff
@@ -810,6 +873,13 @@ class CTransaction : virtual public CTransactionBase
     }
 
     bool ccIsNull() const {
+        return (
+            vcsw_ccin.empty() &&
+            ccoutIsNull()
+        );
+    }
+
+    bool ccoutIsNull() const {
```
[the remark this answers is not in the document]
i-Alex: ok

src/script/interpreter.cpp
i-Alex: right, i will add `if` check first.

src/script/interpreter.cpp
```diff
@@ -1087,9 +1116,23 @@ class CTransactionSignatureSerializer {
  
             if (txTo.IsScVersion() )
             {
+                // Serialize CSW inputs
+                // In case of SIGHASH_ANYONECANPAY:
+                // * if Tx has Sc support version and nIn belongs to the CSW inputs - only the CSW input being signed is serialized
+                // * otherwise skip CSW inputs
+                // Otherwise - serialize all CSW inputs
+                unsigned int nCswInputs = fAnyoneCanPay ?
```
i-Alex, line 1124: same here

src/script/interpreter.cpp
i-Alex: as decided AFK, keep sidechain related data together after regular data.

src/script/interpreter.cpp
i-Alex: to be discussed.

src/txdb.cpp
```diff
@@ -461,3 +462,33 @@ bool CBlockTreeDB::LoadBlockIndexGuts()
 
     return true;
 }
+
+bool CBlockTreeDB::addCertData(const uint256& scId, const int height, const libzendoomc::ScFieldElement &certHash) {
```
i-Alex, line 466: `height` or epoch number?

src/txmempool.cpp
i-Alex: yes

src/coins.cpp
abi87: I think we should prioritize tx spending CSW coins like we do for ordinary txes. The way I understand priority is along the line of what Gavin Andresen explains here (https://www.youtube.com/watch?v=rQ3e1Pzu7iI min 3:39). As of now, I see no reason to do differently

src/primitives/certificate.cpp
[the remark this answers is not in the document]
abi87: Just remove all references to `MakeSignatureChecker` for both txes and certs, within `BITCOIN_TX `blocks and outside of them. I manage to compile this way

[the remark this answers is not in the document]
i-Alex: Great, thanks!

src/txmempool.cpp
abi87: I think the case for `vin` is different from current one, but I happy to discuss.
As per the "two cwss inputs" case, the `remove` over the first cws should correctly remove the containing tx from mapTx and mapSidechain too, hence processing of second cws would be skipped. I think it's important to add a UT confirming this

src/main.cpp
```diff
@@ -2479,11 +2679,18 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
                     LogPrint("cert", "%s():%d - SIDECHAIN-EVENT: failed cancelling scheduled event\n", __func__, __LINE__);
                     return error("DisconnectBlock(): ceasing height cannot be reverted: data inconsistent");
                 }
+                // Remove previously stored DataHash
+                if (!pblocktree->removeCertData(cert.GetScId(), cert.epochNumber))
```
i-Alex, line 2683: Seems, that `pblocktree` is not a proper storage, instead of using it we need to update state (CoinsDB) same as we did for SidechainInfo. And moreover, do the changes firstly to the cache, then do the `flush` for all changes at the same time.
To be discussed.

src/coins.cpp
```diff
@@ -1047,11 +1109,14 @@ bool CCoinsViewCache::HaveScRequirements(const CTransaction& tx, int height)
         const uint256& scId = ft.scId;
         if (HaveSidechain(scId))
         {
-            if (isCeasedAtHeight(scId, height)!= CSidechain::State::ALIVE) {
-                LogPrintf("ERROR: tx[%s] tries to send funds to scId[%s] already ceased at height = %d\n",
-                            txHash.ToString(), scId.ToString(), height);
+            // Q: whe do we not consider the height anymore? Previously we were not able to add FT Tx just before ceasing.
```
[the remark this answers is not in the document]
abi87: As per the height, I believe that it is redundant and misleading and it should be dropped. Reason is that every view is anchored to a block (via setBestBlock) and each block associated with an height (even forks). So somehow the information about which height the view refers to its implicit into it, and should not be passed from outside.
As per the specific case of fwt, I expect no change in behaviour; if there is any, it's a mistake

src/coins.cpp
abi87: Let's discuss this

src/txmempool.cpp
i-Alex: Agreed AFK, add assert and leave a comment why.

src/script/interpreter.cpp
i-Alex: `fHashSingle` usage with CC data must be explored deeply.
In current implementation, when CC outputs are **anyways** serialized, the corner case is the following:
- Tx has fixed set of inputs. Tx has not been signed yet.
- The owner of the first input define the first output and sign the input with `fHashSingle`, so don't care how the change is going to be spent.
- Then Tx was appended with the FT output. It invalidates the signature of the first input, because the message to sign was changed.

Another approach, is when CC outputs are **never** serialized for `fHashNone` and `fHashSingle` flags. The limitation is that we are not able to link the input signature to the CC output, but only to the single output. So no sidechains support for the `fHashSingle`.

We may think about the new signing flags, like "hash single for crosschain output`. But it has it's own problems and implementation complexity.

src/coins.cpp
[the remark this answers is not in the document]
i-Alex: Right, `height` must be removed.
FTs are processed correctly.
I will change the processing of CSWs in the same way.

src/main.cpp
i-Alex: @alsala In our case we do a cast to reference. If failed, it throws bad_cast exception. Do we need to have any additional check here?

src/main.cpp
```diff
@@ -2242,19 +2276,92 @@ bool CheckTxInputs(const CTransactionBase& txBase, CValidationState& state, cons
                 REJECT_INVALID, "bad-txns-inputvalues-outofrange");
     }
 
-    nValueIn += txBase.GetJoinSplitValueIn();
+    for(const CTxCeasedSidechainWithdrawalInput& csw: tx.GetVcswCcIn()) {
```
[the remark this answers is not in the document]
i-Alex: agree.
Do we need to have the same below for joinsplits?

src/primitives/transaction.h
```diff
@@ -349,6 +349,61 @@ class CTxIn
...
+/** An input of a transaction.  It contains the location of the previous
+ * transaction's output that it claims and a signature that matches the
+ * output's public key.
+ */
+class CTxCeasedSidechainWithdrawalInput
+{
+public:
+    CAmount nValue;
+    uint256 scId;
+    int64_t nEpoch;
```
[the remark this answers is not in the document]
i-Alex: right, it must be int32_t

src/rpc/blockchain.cpp
```diff
@@ -15,6 +15,7 @@
 #include "sync.h"
 #include "util.h"
 #include "zen/delay.h"
+#include "txdb.h"
```
i-Alex, line 18: can be removed

src/main.cpp
```diff
@@ -2987,6 +2990,16 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
                                      REJECT_INVALID, "bad-fwd-not-recorded");
                 }
             }
+            
+            for (const CTxCeasedSidechainWithdrawalInput& cswIn:tx.GetVcswCcIn()) {
+                //libzendoomc::ScFieldElement cswIn.nullifier;
```
i-Alex, line 2995: remove comment.

src/main.cpp
```diff
@@ -2987,6 +2990,16 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+                if (!view.GetCswNullifier(cswIn.scId, cswIn.nullifier)) {
+                    view.AddCswNullifier(cswIn.scId, cswIn.nullifier);
+                } else {
+                    return state.DoS(100, error("ConnectBlock(): try to use existed nullifier Tx [%s]", tx.GetHash().ToString()),
+                             REJECT_INVALID, "bad-txns-csw-input-not-applicable");
```
i-Alex, line 3000: let's define another message, like "bad-txns-csw-input-nullifier"

src/main.cpp
```diff
@@ -2687,7 +2687,10 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
                     LogPrint("cert", "%s():%d ApplyTxInUndo returned FALSE on tx [%s] \n", __func__, __LINE__, tx.GetHash().ToString());
                     fClean = false;
                 }
+            }
 
+            for (const CTxCeasedSidechainWithdrawalInput& cswIn:tx.GetVcswCcIn()) {
```
i-Alex, line 2692: move this before
`for (const CTxForwardTransferOut& fwdTransfer: tx.GetVftCcOut()) {
    if (!view.CancelSidechainEvent(fwdTransfer, pindex->nHeight)) {`
to keep exactly reversed order to ConnectBlock

src/coins.h
```diff
@@ -283,6 +283,24 @@ class CCoinsKeyHasher
...
+public:
+    CCswNullifiersKeyHasher();
+
+    /**
+     * This *must* return size_t. With Boost 1.46 on 32-bit systems the
+     * unordered_map will behave unpredictably if the custom hasher returns a
+     * uint64_t, resulting in failures when syncing the chain (#4634).
+     */
+    size_t operator()(const std::pair<uint256, libzendoomc::ScFieldElement>& key) const {
+        return key.second.GetHash(salt);
```
i-Alex, line 300: why we consider only `key.second` hash? what about `key.first`?

src/coins.h
```diff
@@ -411,6 +445,10 @@ class CCoinsView
 
     //! Get the current "tip" or the latest anchored tree root in the chain
     virtual uint256 GetBestAnchor() const;
+    
+    //! Retrieve existance of CSW nullifier for specified Sidechain.
+    virtual bool GetCswNullifier(const uint256& scId,
```
i-Alex, line 450: `GetCswNullifier` -> `HaveCswNullifier`

src/coins.cpp
```diff
@@ -503,13 +512,47 @@ void CCoinsViewCache::SetBestBlock(const uint256 &hashBlockIn) {
...
+    ret.first->second.flags |= CCswNullifiersCacheEntry::DIRTY;
+}
+
+bool CCoinsViewCache::GetCswNullifier(const uint256& scId,
+                         const libzendoomc::ScFieldElement &nullifier) const {
+    std::pair<uint256, libzendoomc::ScFieldElement> position = std::make_pair(scId, nullifier);
+    
+    CCswNullifiersMap::iterator it = cacheCswNullifiers.find(position);
+    if (it != cacheCswNullifiers.end())
+        return true;
```
i-Alex, line 537: instead of `return true` -> check the `it` flag value.
It may be the case, when in `base` the nullifer was present, but in the `cache` we removed it. So, cache should return `false`, because iterator `flag == ERASED`.

src/coins.cpp
```diff
@@ -647,6 +690,26 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins,
     }
     mapSidechainEvents.clear();
 
+    for (CCswNullifiersMap::iterator child_it = cswNullifiers.begin(); child_it != cswNullifiers.end();)
+    {
+        if (child_it->second.flags & CNullifiersCacheEntry::DIRTY) { // Ignore non-dirty entries (optimization).
+            CCswNullifiersMap::iterator parent_it = cswNullifiers.find(child_it->first);
```
i-Alex, line 696: not `cswNullifiers` but `cacheCswNullifiers`

src/coins.cpp
```diff
@@ -647,6 +690,26 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins,
...
+
+            if (parent_it == cswNullifiers.end()) {
```
i-Alex, line 698: same here.

i-Alex: Honestly, this `for` loop can be simplified a lot.
Do similar to mapSidechains update:
Loop thought `cswNullifiers`. If item flag is Fresh or Erased -> do checks/asserts and update the value in `cacheCswNullifiers` map.
If item flag is `DEFAULT` -> do nothing. `default: assert(false)`

src/rpc/blockchain.cpp
```diff
@@ -1503,27 +1502,25 @@ UniValue getscgenesisinfo(const UniValue& params, bool fHelp)
...
-            "  \"data\":            xx,      (string) existance of nullifier\n"
+            "  \"data\":            xx,      (bool) existance of nullifier\n"
             "}\n"
 
-            /*"\nExamples\n"
-            + HelpExampleCli("getscinfo", "\"1a3e7ccbfd40c4e2304c3215f76d204e4de63c578ad835510f580d529516a874\"")
-            + HelpExampleCli("getscinfo", "\"*\" true false 2 10")
-            + HelpExampleCli("getscinfo", "\"*\" ")*/
+            "\nExamples\n"
+            + HelpExampleCli("getscinfo", "\"1a3e7ccbfd40c4e2304c3215f76d204e4de63c578ad835510f580d529516a874\""
```
i-Alex, line 1520: "getscinfo" -> "checkcswnullifier"

src/txdb.cpp
```diff
@@ -264,11 +280,39 @@ bool CCoinsViewDB::BatchWrite(CCoinsMap &mapCoins,
     
     for (CCswNullifiersMap::iterator it = cswNullifies.begin(); it != cswNullifies.end();) {
         const std::pair<uint256, libzendoomc::ScFieldElement>& position = it->first;
-        BatchWriteCswNullifier(batch, position.first, position.second, it->second.entered);
+        BatchWriteCswNullifier(batch, position.first, position.second, (it->second.flags == CCswNullifiersCacheEntry::FRESH));
```
i-Alex, line 283: Pass as a third parameter `it->second` directly. Inside, do `switch(it->second.flag)` and process `DEFAULT` case as well.
Now, you will remove the entry even if the status was `DEFAULT`. See, `BatchSidechains` method,

src/coins.cpp
```diff
@@ -690,25 +724,54 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins,
...
+        switch (entryToWrite.second.flags) {
+            case CCertDataHashCacheEntry::Flags::FRESH:
+                assert(
+                    itLocalCacheEntry == cacheCertDataHashes.end() ||
+                    itLocalCacheEntry->second.flags == CCertDataHashCacheEntry::Flags::ERASED
+                ); //A fresh entry should not exist in localCache or be already erased
+                cacheCertDataHashes[entryToWrite.first] = entryToWrite.second;
+                break;
+            case CCertDataHashCacheEntry::Flags::DIRTY: //A dirty entry may or may not exist in localCache
+                cacheCertDataHashes[entryToWrite.first] = entryToWrite.second;
```
i-Alex, line 760: @abi87 What if in the cache some entry was marked as `FRESH`, but in the `entryToWrite` it's `DIRTY`?
So the entry was created initially in the first layer cache, and then modified in the second layer cache.
I think, that we need to leave in the `cacheCertDataHashes[entryToWrite.first]` flag value as is, ignoring the flag in the `entryToWrite.second`.
PS: though the current code in txdb processes `FRESH` and `DIRTY` in the same way, but we need to be more careful in the cache level.
What do you think?

src/txdb.cpp
```diff
@@ -264,11 +280,39 @@ bool CCoinsViewDB::BatchWrite(CCoinsMap &mapCoins,
...
+        std::pair<libzendoomc::ScFieldElement, libzendoomc::ScFieldElement> prevData;
+        libzendoomc::ScFieldElement cumulativeCertData;
+
+        switch(it->second.flags) {
+            case CCertDataHashCacheEntry::Flags::FRESH:
+            case CCertDataHashCacheEntry::Flags::DIRTY:
+                if (GetCertData(scId, epochId - 1, prevData)) {
+                    cumulativeCertData = calculateCumulativeCertDataHash(prevData.first, prevData.second);
+                }
+                BatchWriteCertDataHash(batch, scId, epochId, it->second.certDataHash, cumulativeCertData);
```
i-Alex, line 302: I have a doubt that we do `cumulativeCertData` computation here (which also requires `GetCertData` read operation during the `BatchWrite`. As for me, this is a part of business logic that we need to put to cache method `AddCertHashData`. So, the `cumulativeCertData` should be a part of `CCertDataHashCacheEntry`.

Moreover, we don't need to recalculate `CCertDataHashCacheEntry` in case if there is no epoch switch compare to the last cert data stored.

src/txdb.cpp
```diff
@@ -264,11 +280,39 @@ bool CCoinsViewDB::BatchWrite(CCoinsMap &mapCoins,
...
+        BatchWriteCswNullifier(batch, position.first, position.second, (it->second.flags == CCswNullifiersCacheEntry::FRESH));
         CCswNullifiersMap::iterator itOld = it++;
         cswNullifies.erase(itOld);
     }
 
+    // Store cert Data Hashes
+    for (CCertDataHashMap::iterator it = certDataHashes.begin(); it != certDataHashes.end();) {
+        const std::pair<uint256, int>& position = it->first;
+        uint256 scId = position.first;
+        int epochId = position.second;
```
i-Alex, line 292: epochNumber

src/txdb.cpp
```diff
@@ -264,11 +280,39 @@ bool CCoinsViewDB::BatchWrite(CCoinsMap &mapCoins,
...
+        std::pair<libzendoomc::ScFieldElement, libzendoomc::ScFieldElement> prevData;
+        libzendoomc::ScFieldElement cumulativeCertData;
+
+        switch(it->second.flags) {
+            case CCertDataHashCacheEntry::Flags::FRESH:
+            case CCertDataHashCacheEntry::Flags::DIRTY:
+                if (GetCertData(scId, epochId - 1, prevData)) {
+                    cumulativeCertData = calculateCumulativeCertDataHash(prevData.first, prevData.second);
```
i-Alex, line 300: Let's put the comment somewhere, that we store cumulative cert data hash up to the `epochNumber` (so without the current epoch `certDataHash`.

src/coins.cpp
```diff
@@ -1562,6 +1664,9 @@ bool CCoinsViewCache::RestoreScInfo(const CScCertificate& certToRevert, const CS
     scIt->second.scInfo.prevBlockTopQualityCertQuality         = sidechainUndo.prevTopCommittedCertQuality;
     scIt->second.scInfo.prevBlockTopQualityCertBwtAmount       = sidechainUndo.prevTopCommittedCertBwtAmount;
 
+    assert(sidechainUndo.contentBitMask & CSidechainUndoData::AvailableSections::CERT_DATA_HASH);
```
i-Alex, line 1667: not true if we rollback the first cert for given epoch.

src/sc/proofverifier.cpp
```diff
@@ -74,6 +74,48 @@ namespace libzendoomc{
...
+            LogPrint("zendoo_mc_cryptolib",
+                    "%s():%d - failed to deserialize \"constant\": %s \n",
+                    __func__, __LINE__, ToString(zendoo_get_last_error()));
+            zendoo_clear_error();
+
+            zendoo_field_free(prevCumulativeHashInput);
+            return false;
+        }
+
+        const field_t* hashInput[] = {prevCumulativeHashInput, prevCertHashInput};
```
i-Alex, line 106: possible memory leak?

src/coins.cpp
```diff
@@ -864,21 +1136,106 @@ int CSidechain::EpochFor(int targetHeight) const { return CScCertificate::EPOCH_
...
+                __func__, __LINE__, tx.ToString(), csw.ToString() );
+            return state.Invalid(error("received a csw for sc without CSW support"),
+                 REJECT_INVALID, "tx-csw-input-sc-without-csw-support");
+        }
+
+        // add a new balance entry in the map or increment it if already there
+        cswTotalBalances[csw.scId] += csw.nValue;
+
+        for(const CTxCeasedSidechainWithdrawalInput& cswIn : tx.GetVcswCcIn()) {
+            if (pcoinsTip->HaveCswNullifier(csw.scId, cswIn.nullifier)) {
```
alsala, line 1200: it should not be calle on pcoinsTip, but on this instance instead.

src/coins.cpp
```diff
@@ -503,13 +539,149 @@ void CCoinsViewCache::SetBestBlock(const uint256 &hashBlockIn) {
...
+    ret.first->second.flag = CCswNullifiersCacheEntry::ERASED;
+}
+
+bool CCoinsViewCache::HaveCswNullifier(const uint256& scId,
+                         const libzendoomc::ScFieldElement &nullifier) const {
+    std::pair<uint256, libzendoomc::ScFieldElement> position = std::make_pair(scId, nullifier);
+    
+    CCswNullifiersMap::iterator it = cacheCswNullifiers.find(position);
+    if (it != cacheCswNullifiers.end())
+        return (it->second.flag = CCswNullifiersCacheEntry::ERASED);
```
alsala, line 562: This should be !=, that is return false if an entry there and is not flagged as ERASED

src/coins.cpp
```diff
@@ -503,13 +539,149 @@ void CCoinsViewCache::SetBestBlock(const uint256 &hashBlockIn) {
...
+
+    bool tmp = base->HaveCswNullifier(scId, nullifier);
+
+    if (tmp) {
+        CCswNullifiersCacheEntry entry;
+        entry.flag = CCswNullifiersCacheEntry::DEFAULT;
+        cacheCswNullifiers.insert(std::make_pair(position, entry));
```
alsala, line 569: the return value should be checked: if ret.second == false it means the insertion did not take place. Anyway why overwrite the entry?

src/main.cpp
alsala: We should add a try/catch block, since when dynamic_cast fails for a reference type, it throws a 'bad_cast' exception

src/main.cpp
```diff
@@ -2242,6 +2176,15 @@ bool CheckTxInputs(const CTransactionBase& txBase, CValidationState& state, cons
                 REJECT_INVALID, "bad-txns-inputvalues-outofrange");
     }
 
+    try {
+        nValueIn += txBase.GetCSWValueIn();
+        if (!MoneyRange(nValueIn))
+            throw std::runtime_error("Total inputs value out of range.");
```
alsala, line 2182: We should not rethrow the std::runtime_error(), but should do return state.DoS instead as in the block below

src/rpc/blockchain.cpp
```diff
@@ -1494,6 +1502,64 @@ UniValue getscgenesisinfo(const UniValue& params, bool fHelp)
 
 }
 
+UniValue checkcswnullifier(const UniValue& params, bool fHelp)
+{
+    if (fHelp || params.size() == 0 || params.size() > 2)
```
alsala, line 1507: if both params are mandatory, it should be if (fHelp || params.size != 2)

src/rpc/blockchain.cpp
```diff
@@ -1164,6 +1169,8 @@ bool FillScRecordFromInfo(const uint256& scId, const CSidechain& info, CSidechai
                 sc.push_back(Pair("unconf wCertVk", HexStr(info.creationData.wCertVk)));
                 sc.push_back(Pair("unconf customData", HexStr(info.creationData.customData)));
                 sc.push_back(Pair("unconf constant", HexStr(info.creationData.constant)));
+                if(info.creationData.wCeasedVk.is_initialized())
+                    sc.push_back(Pair("wCeasedVk", HexStr(info.creationData.wCeasedVk.get())));
```
albertog78, line 1173: add else as managed for MBTRs

albertog78 (changes requested):

src/txdb.cpp
```diff
@@ -241,6 +287,35 @@ bool CCoinsViewDB::BatchWrite(CCoinsMap &mapCoins,
...
+        const std::pair<uint256, libzendoomc::ScFieldElement>& position = it->first;
+        BatchWriteCswNullifier(batch, position.first, position.second, it->second);
+        CCswNullifiersMap::iterator itOld = it++;
+        cswNullifies.erase(itOld);
+    }
+
+    // Store cert Data Hashes
+    for (CCertDataHashMap::iterator it = certDataHashes.begin(); it != certDataHashes.end();) {
+        const std::pair<uint256, int>& position = it->first;
+        uint256 scId = position.first;
```
alsala, line 301: const & would avoid copying

src/wallet/rpcwallet.cpp
```diff
@@ -783,7 +780,7 @@ static void ScHandleTransaction(CWalletTx& wtx, std::vector<CcRecipientVariant>&
     if (!pwalletMain->CommitTransaction(wtx, keyChange))
         throw JSONRPCError(RPC_WALLET_ERROR, "Transaction commit failed");
 }
```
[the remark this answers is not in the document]
alsala: An issue ha s been open:
 https://github.com/HorizenOfficial/zend_oo/issues/86

abi87 (changes requested):

src/primitives/transaction.cpp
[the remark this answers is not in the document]
abi87: Am I not going blind?

abi87 (changes requested):

src/sc/proofverifier.cpp
```diff
@@ -65,6 +74,48 @@ namespace libzendoomc{
...
     }
 
+    ScFieldElement CalculateCertDataHash(const CScCertificate& cert) {
+        // TODO Replace with hash of merkel tree supplied by SC.
+        ScFieldElement certDataHash;
+        certDataHash.SetHex(cert.GetHash().GetHex());
+        return certDataHash;
+    }
+
+    bool CalculateCumulativeCertDataHash(const ScFieldElement& prevCumulativeHash, const ScFieldElement& prevCertHash, ScFieldElement& CumulativeHashOutput) {
```
abi87, line 84: Could we return the ScFieldElement CumulativeHashOutput instead of bool and test outside if it is valid or not?
I personally hate this assembly-like style; but I can live with it

abi87 (changes requested):

src/coins.cpp
```diff
@@ -223,14 +230,41 @@ bool CCoinsViewBacked::CheckQuality(const CScCertificate& cert)
...
+                                                                                            certDataHashes); }
 bool CCoinsViewBacked::GetStats(CCoinsStats &stats)                                  const { return base->GetStats(stats); }
 
 CCoinsKeyHasher::CCoinsKeyHasher() : salt(GetRandHash()) {}
+CCswNullifiersKeyHasher::CCswNullifiersKeyHasher() : salt() {GetRandBytes(reinterpret_cast<unsigned char*>(salt), BUF_LEN);}
+
+size_t CCswNullifiersKeyHasher::operator()(const std::pair<uint256, libzendoomc::ScFieldElement>& key) const {
+    uint32_t buf[BUF_LEN/4];
+    memcpy(buf, key.first.begin(), 32);
+    memcpy((buf + 8), key.second.begin(), SC_FIELD_SIZE);
```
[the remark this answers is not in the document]
abi87: With the help of @alsala  I understood that there is no overflow.
I would really really suggest to:

1. Set `static const size_t BUF_LEN = (sizeof(uint256) + SC_FIELD_SIZE)/sizeof(uint32_t);` in declaration
2. Set `uint32_t salt[BUF_LEN/4];` in declaration
3. Set `    memcpy(buf, key.first.begin(), sizeof(uint256));
    memcpy((buf + sizeof(uint256)/sizeof(uint32_t)), key.second.begin(), SC_FIELD_SIZE);` in operator
4. Discuss deeply about performances. For reference check how much Bitcoin core devs have worked on similar issue: https://github.com/bitcoin/bitcoin/pull/4413

src/main.cpp
```diff
@@ -2315,6 +2398,64 @@ bool ContextualCheckInputs(const CTransactionBase& tx, CValidationState &state,
     return true;
 }
 
+bool ContextualCheckCertInputs(const CScCertificate& cert, CValidationState &state, const CCoinsViewCache &inputs, bool fScriptChecks, const CChain& chain, unsigned int flags, bool cacheStore, const Consensus::Params& consensusParams, std::vector<CScriptCheck> *pvChecks)
```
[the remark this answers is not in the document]
i-Alex: done

src/primitives/certificate.cpp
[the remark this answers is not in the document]
i-Alex: done

src/primitives/transaction.cpp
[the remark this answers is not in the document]
i-Alex: done

src/txmempool.cpp
i-Alex: done

src/main.cpp
```diff
@@ -1450,6 +1462,15 @@ bool AcceptTxToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTran
             }
         }
 
+        // Check if this tx does CSW with the nullifier already present in the mempool
+        for(const CTxCeasedSidechainWithdrawalInput& csw: tx.GetVcswCcIn())
+        {
+            if (pool.mapSidechains.count(csw.scId) != 0 && pool.mapSidechains.at(csw.scId).cswNullifiers.count(csw.nullifier) != 0) {
```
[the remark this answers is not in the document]
i-Alex: done

src/main.cpp
```diff
@@ -2309,6 +2416,99 @@ bool ContextualCheckInputs(const CTransactionBase& tx, CValidationState &state,
                     return state.DoS(100,false, REJECT_INVALID, strprintf("mandatory-script-verify-flag-failed (%s)", ScriptErrorString(check.GetScriptError())));
                 }
             }
+
+            unsigned int vinSize = tx.GetVin().size();
```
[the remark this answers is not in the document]
i-Alex: done

src/primitives/transaction.h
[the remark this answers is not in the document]
i-Alex: done

src/bitcoin-tx.cpp
```diff
@@ -434,6 +434,42 @@ static void MutateTxSign(CMutableTransaction& tx, const string& flagStr)
             fComplete = false;
     }
 
+    if(mergedTx.IsScVersion())
```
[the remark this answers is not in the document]
i-Alex: ok

src/coins.cpp
i-Alex: done

src/coins.cpp
i-Alex: done

src/rpc/blockchain.cpp
i-Alex: done

src/main.cpp
```diff
@@ -1494,16 +1415,23 @@ bool AcceptTxToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTran
...
+                return state.Invalid(error("%s(): sidechain is redeclared or coins are forwarded to unknown/ceased sidechain", __func__),
                                     REJECT_INVALID, "bad-sc-tx");
             }
  
             // are the joinsplit's requirements met?
             if (!view.HaveJoinSplitRequirements(tx))
                 return state.Invalid(error("%s(): joinsplit requirements not met", __func__),
                                      REJECT_DUPLICATE, "bad-txns-joinsplit-requirements-not-met");
+
+            auto scVerifier = disconnecting ? libzendoomc::CScProofVerifier::Strict() : libzendoomc::CScProofVerifier::Disabled();
```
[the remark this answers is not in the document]
i-Alex: done

src/script/interpreter.cpp
i-Alex: not done

src/main.cpp
```diff
@@ -2148,6 +2170,16 @@ bool CheckTxInputs(const CTransactionBase& txBase, CValidationState& state, cons
                 REJECT_INVALID, "bad-txns-inputvalues-outofrange");
     }
 
+    try {
+        nValueIn += txBase.GetCSWValueIn();
+        if (!MoneyRange(nValueIn))
+            return state.DoS(100, error("CheckInputs(): Total inputs value out of range."),
+                REJECT_INVALID, "bad-txns-inputvalues-outofrange");
+    } catch (const std::runtime_error& e) {
```
abi87, line 2178: I think you do not need the try/catch here currently. There is no cast or anything. Remove it

src/main.cpp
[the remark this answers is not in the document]
abi87: Ok with just a comment on a unnecessary try/catch.

abi87 (changes requested):
A try-catch can be eliminated. Scratch that, I was wrong

src/main.cpp
abi87: I do not see where the first runtime_error is thrown in the first place

src/coins.cpp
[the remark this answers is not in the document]
abi87: Points 1/2/3 addressed. I'd leave point 4 for final discussion

src/uint256.cpp
```diff
@@ -149,3 +156,80 @@ uint64_t uint256::GetHash(const uint256& salt) const
...
+    a += pn[21] ^ salt_pn[21];
+    b += pn[22] ^ salt_pn[22];
+    c += pn[23] ^ salt_pn[23];
+    
+    HashFinal(a, b, c);
+
+    return ((((uint64_t)b) << 32) | c);
+}
+
+uint64_t CalculateHash(const uint32_t* const src, size_t length, const uint32_t* const salt)
```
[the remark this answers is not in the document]
abi87: There must be faster solutions

src/primitives/certificate.cpp
```diff
@@ -232,6 +237,34 @@ std::shared_ptr<const CTransactionBase> CScCertificate::MakeShared() const
 }
 #else
 
+bool CScCertificate::ContextualCheckInputs(CValidationState &state, const CCoinsViewCache &view, bool fScriptChecks,
+          const CChain& chain, unsigned int flags, bool cacheStore, const Consensus::Params& consensusParams,
+          std::vector<CScriptCheck> *pvChecks) const
+{
+    return ::ContextualCheckCertInputs(*this, state, view, fScriptChecks, chain, flags, cacheStore, consensusParams, pvChecks);
```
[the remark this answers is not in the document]
abi87: I took the liberty of removing this and handle slightly differently the miner. Please review @i-Alex , @alsala

src/script/interpreter.cpp
albertog78: REACHED THIS POINT

albertog78 (changes requested):

src/txmempool.cpp
```diff
@@ -715,6 +793,25 @@ void CTxMemPool::removeStaleTransactions(const CCoinsViewCache * const pCoinsVie
...
+        {
+            if(pCoinsView->GetSidechainState(csw.scId) != CSidechain::State::CEASED)
+            {
+                txesToRemove.insert(tx.GetHash());
+                continue;
+            }
+        }
+    }
+
+    for(const auto& hash: txesToRemove)
```
albertog78, line 807: this is already done at line 826

src/txmempool.cpp
```diff
@@ -915,11 +1017,35 @@ void CTxMemPool::check(const CCoinsViewCache *pcoins) const
...
+                assert(pcoins->GetSidechainState(fwd.scId) == CSidechain::State::ALIVE);
+        }
+
+        std::map<uint256, CAmount> cswBalances;
+        for(const CTxCeasedSidechainWithdrawalInput& csw: tx.GetVcswCcIn()) {
+            //CSW must be duly recorded in mapSidechain
+            assert(mapSidechains.count(csw.scId) != 0);
+            const auto& cswNullifierPos = mapSidechains.at(csw.scId).cswNullifiers.find(csw.nullifier);
+            assert(cswNullifierPos != mapSidechains.at(csw.scId).cswNullifiers.end());
+
```
albertog78, line 1033: assert csw nullifier map value matches tx hash

src/uint256.cpp
```diff
@@ -156,3 +160,35 @@ uint64_t uint256::GetHash(const uint256& salt) const
...
+        b += pn[1] ^ salt_pn[1];
+        c += pn[2] ^ salt_pn[2];
+        HashMix(a, b, c);
+
+        length -= 3;
+        pn += 3;
+        salt_pn += 3;
+    }
+
+    switch(length) {
```
albertog78, line 182: transform the switch without "breaks" into something more readable. Unit test this comparing results with GetHash where possible

albertog78 (changes requested):

abi87 (approved):

albertog78 (approved):

PR #80, opened 2021-01-15
Modified fundrawtransaction cmd in order to support Ceased Sc withdra…

alsala:
…wal req in tx

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1805,7 +1805,10 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived, list<COutputEntry>&
         for (const JSDescription & js : wrappedTx.GetVjoinsplit()) {
             nValueIn += js.vpub_new;
         }
-        nFee = nDebit - nValueOut + nValueIn;
+        // nDebit has only vin contribution, we must add the ceased sc with part if any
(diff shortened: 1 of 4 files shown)


src/wallet/wallet.cpp
```diff
@@ -3148,12 +3151,19 @@ bool CWallet::FundTransaction(CMutableTransaction& tx, CAmount &nFeeRet, int& nC
     
     CCoinControl coinControl;
     coinControl.fAllowOtherInputs = true;
-    BOOST_FOREACH(const CTxIn& txin, tx.vin)
+    for (const CTxIn& txin: tx.vin)
         coinControl.Select(txin.prevout);
 
+    // if we have ceased sc withdrawl inputs, they must be taken into account when doing computations for funding input amount
+    CAmount cswInTotAmount = 0;
```
i-Alex, line 3158: `CAmount cswTotAmount = tx.GetCSWValueIn()` instead of the loop.

i-Alex (changes requested):

src/wallet/wallet.cpp
alsala: tx is a mutable obj and does not have such method

i-Alex: right.

i-Alex (approved):

abi87 (approved):

PR #81, opened 2021-01-19
Fix for getting correct sc state; removed list of ptr when doing a re…

alsala:
…move call in mempool

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2093,7 +2093,7 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
         pcoinsTip->GetScIds(allScIds);
         for(const auto& scId: allScIds)
         {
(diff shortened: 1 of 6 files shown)


src/wallet/wallet.cpp
```diff
@@ -2093,7 +2093,7 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
         pcoinsTip->GetScIds(allScIds);
         for(const auto& scId: allScIds)
         {
-            if (pcoinsTip->isCeasedAtHeight(scId, chainActive.Height()) != CSidechain::State::ALIVE) {
```
i-Alex, line 2096: Do we need `isCeasedAtHeight` method at all? Seems we always need to use `GetSidechainState`

alsala: No, you are right. It is only in UT for the time being we chose to leave it there, but we might as well remove it.

alsala: Added a comment in the header for that

i-Alex (approved):

i-Alex (approved):

PR #82, opened 2021-01-19
Introduction of ad-hoc type for ScTxsCommitment

abi87:
Currently ScTxsCommitment is a uint256 type. This will change soon and probably multiple times. These PR introduce a new type for ScTxsCommitment. It is called CFakePoseidonHash, it simply wraps uint256, introducing an interface which should make subsequent changes to ScTxsCommitment much simpler.

Currently the PR targets sidechain_dev

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -183,7 +183,7 @@ public:
     //! block header
     int nVersion;
     uint256 hashMerkleRoot;
-    uint256 hashScTxsCommitment;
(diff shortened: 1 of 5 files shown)


alsala (approved):

abi87:
Closed since superseeded by https://github.com/HorizenOfficial/zend_oo/pull/89 and https://github.com/HorizenOfficial/zend_oo/pull/83

PR #89, opened 2021-02-12
Sc commitment tree cumulative hash

alsala:


src/main.cpp
```diff
@@ -3871,6 +3871,18 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block)
     if(pindexNew->nChainDelay != 0) {
         LogPrintf("%s: Block belong to a chain under punishment Delay VAL: %i BLOCKHEIGHT: %d\n",__func__, pindexNew->nChainDelay,pindexNew->nHeight);
     }
+
+    if (pindexNew->pprev && pindexNew->nVersion == BLOCK_VERSION_SC_SUPPORT )
+    {
+        const CPoseidonHash& prevScCumTreeHash = (pindexNew->pprev->nVersion == BLOCK_VERSION_SC_SUPPORT) ? pindexNew->pprev->scCumTreeHash : CPoseidonHash();
+        pindexNew->scCumTreeHash = CPoseidonHash::ComputeHash(prevScCumTreeHash, block.hashScTxsCommitment); 
+    }
+    else
```
i-Alex, line 3880: Do we need else case at all?

src/miner.cpp
```diff
@@ -731,7 +731,7 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn,  unsigned int nBlo
 
         // Fill in header
         pblock->hashPrevBlock  = pindexPrev->GetBlockHash();
-        pblock->hashScTxsCommitment = pblock->BuildScTxsCommitment();
+        pblock->hashScTxsCommitment = CPoseidonHash(pblock->BuildScTxsCommitment());
```
i-Alex, line 734: `pblock->BuildScTxsCommitment()` may return `CPoseidonHash` type.

i-Alex (changes requested):

src/coins.cpp
```diff
@@ -946,9 +939,26 @@ bool CCoinsViewCache::IsCertApplicableToState(const CScCertificate& cert, int nH
...
+    CPoseidonHash scCumTreeHash_start = prev_start_epoch_block_index->scCumTreeHash;
+    const libzendoomc::ScFieldElement& scCumTreeHash_start_Fe = scCumTreeHash_start.GetFieldElement();
+
+    CPoseidonHash scCumTreeHash_end   = prev_end_epoch_block_index->scCumTreeHash;
+    const libzendoomc::ScFieldElement& scCumTreeHash_end_Fe = scCumTreeHash_end.GetFieldElement();
+
+    uint256 prev_end_epoch_block_hash = prev_end_epoch_block_index->GetBlockHash();
 
     // Verify certificate proof
     if (!scVerifier.verifyCScCertificate(scInfo.creationData.constant, scInfo.creationData.wCertVk, prev_end_epoch_block_hash, cert)){
```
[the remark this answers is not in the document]
i-Alex: Add TODOs to remove `cert.endEpochBlockHash` field from the Certificate and all related checks as well.

src/coins.cpp
```diff
@@ -941,25 +942,27 @@ bool CCoinsViewCache::IsCertApplicableToState(const CScCertificate& cert, int nH
 
 
     // Retrieve previous end epoch block info for certificate proof verification
-    int prev_start_epoch_block_height = scInfo.StartHeightForEpoch(cert.epochNumber-1);
-    int prev_end_epoch_block_height   = scInfo.StartHeightForEpoch(cert.epochNumber) - 1;
+    int prev_end_epoch_block_height = scInfo.StartHeightForEpoch(cert.epochNumber) - 1;
+    int curr_end_epoch_block_height = scInfo.StartHeightForEpoch(cert.epochNumber + 1) - 1;
```
i-Alex, line 946: Minor improvement. We may introduce `EndHeightForEpoch` method to avoid every time doing like `scInfo.StartHeightForEpoch(cert.epochNumber + 1) - 1)`
So, instead here and in a lot of other places we may type:
`scInfo.EndHeightForEpoch(cert.epochNumber)`
Seems much more clear to me.

`EndHeightForEpoch` may reuse the logic of `StartHeightForEpoch`
What do you think @alsala @VolodymyrBabaiev ?

alsala: Ok for me, it is useful

VolodymyrBabaiev: Yes, make sense.

src/miner.cpp
VolodymyrBabaiev: CPoseidonHash is a temporary wrapper until 256-bit Poseidon Hash will be introduced.

src/main.cpp
```diff
@@ -3025,7 +3025,7 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     {
         const uint256& scTxsCommittment = scCommitmentBuilder.getCommitment();
 
-        if (block.hashScTxsCommitment != scTxsCommittment)
+        if (block.hashScTxsCommitment != CPoseidonHash(scTxsCommittment))
```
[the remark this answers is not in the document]
VolodymyrBabaiev: CPoseidonHash is a temporary wrapper until 256-bit Poseidon Hash will be introduced.

i-Alex (approved):

VolodymyrBabaiev (approved):

i-Alex (approved):

src/gtest/test_cumulativehash.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+    CBlockIndex originalpindex;
+    originalpindex.scCumTreeHash = CPoseidonHash(libzendoomc::ScFieldElement({std::vector<unsigned char>(size_t(SC_FIELD_SIZE), 'a')}));
+
+    CDataStream ssValue(SER_DISK, PROTOCOL_VERSION);
+    ssValue << CDiskBlockIndex(&originalpindex);
+    CDiskBlockIndex diskpindex;
+    ssValue >> diskpindex;
+
+    EXPECT_TRUE(originalpindex.scCumTreeHash == diskpindex.scCumTreeHash);
```
abi87, line 26: I really like the structure of this test. It makes it precise since you compare the whole original/retrieved objects with == but it's also robust since you do not depend on internal structure of serialization!

src/main.cpp
```diff
@@ -3737,6 +3737,13 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block)
     if(pindexNew->nChainDelay != 0) {
         LogPrintf("%s: Block belong to a chain under punishment Delay VAL: %i BLOCKHEIGHT: %d\n",__func__, pindexNew->nChainDelay,pindexNew->nHeight);
     }
+
+    if (pindexNew->pprev && pindexNew->nVersion == BLOCK_VERSION_SC_SUPPORT )
+    {
+        const CPoseidonHash& prevScCumTreeHash = (pindexNew->pprev->nVersion == BLOCK_VERSION_SC_SUPPORT) ? pindexNew->pprev->scCumTreeHash : CPoseidonHash();
+        pindexNew->scCumTreeHash = CPoseidonHash::ComputeHash(prevScCumTreeHash, block.hashScTxsCommitment); 
+    }
+
```
abi87, line 3746: If/else would have been appreciated, but logic seems fine to me

abi87 (changes requested):

src/gtest/test_cumulativehash.cpp
VolodymyrBabaiev: Thank you!

src/coins.cpp
```diff
@@ -1076,10 +1067,28 @@ bool CCoinsViewCache::IsCertApplicableToState(const CScCertificate& cert, libzen
...
+    int curr_end_epoch_block_height = sidechain.GetEndHeightForEpoch(cert.epochNumber);
 
+    CBlockIndex* prev_end_epoch_block_index = chainActive[prev_end_epoch_block_height];
+    CBlockIndex* curr_end_epoch_block_index = chainActive[curr_end_epoch_block_height];
+
+    assert(prev_end_epoch_block_index);
+    assert(curr_end_epoch_block_index);
+
+    // TODO check that ithose are the correct heights for both scCumTreeHash objs
+    // TODO add the cumulative committment tree hash objs to the verify call
```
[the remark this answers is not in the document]
abi87: Actually, given the lengthy integration process we have undertaken, it proved useful having these kind of placeholders

abi87 (approved):

src/chain.cpp
```diff
@@ -14,6 +13,9 @@ using namespace std;
  * CChain implementation
  */
 
+const CFieldElement CBlockIndex::defaultScCumTreeHash =
```
albertog78, line 16: Introduce a utility method to retrieve from zendoo_mc_cryptolib a constant phantom hash field element and use it everywhere it is needed a constant value whose preimage has to be unknown

src/chain.h
```diff
@@ -214,11 +219,13 @@ class CBlockIndex
...
         nVersion       = 0;
         hashMerkleRoot = uint256();
-        hashScTxsCommitment = uint256();
+        hashScTxsCommitment.SetNull();
         nTime          = 0;
         nBits          = 0;
         nNonce         = uint256();
         nSolution.clear();
+
+        scCumTreeHash = CBlockIndex::defaultScCumTreeHash;
```
albertog78, line 228: Why not SetNull()?

albertog78 (changes requested):

albertog78 (approved):

PR #101, opened 2021-03-20
Cleanup zendoo-mc-crypto lib integration points

abi87:


src/sc/sidechaintypes.h
```diff
@@ -10,15 +10,47 @@
...
+    bool operator==(const CZendooCctpObject& rhs) const { return this->byteVector == rhs.byteVector; }
+    bool operator!=(const CZendooCctpObject& rhs) { return !(*this == rhs); }
+
+    // SERIALIZATION SECTION
+    ADD_SERIALIZE_METHODS
+    template <typename Stream, typename Operation>
+    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
+        READWRITE(byteVector);
+        //Allow zero-length byteVector for, e.g. serialization/hashing of empty txes/certs
+        if (byteVector.size() != 0 && byteVector.size() != SerializedSize())
```
i-Alex, line 43: I'm worried about using byte vector to serialize a constant size value. For example, for FieldElement we always know the size. So by default, uninitialized/invalid FE can be set to the static value of `FF...FF` - which is invalid FE.
We may make FE backed by base_blob of given size that will save us some bytes during serialization: no `VarInt` serialization for byte array length.
Current approach, as for me, contradicts with the regular serialization approach and with how we define an "uninitialized" state: setting empty byte vector is like setting a different invalid type instead of setting an invalid value specific for the type (-1 for Amount, or "FF...FF" for FE).
Maybe I'm missing something, but hope it's useful.

src/sc/sidechaintypes.h
```diff
@@ -29,96 +61,104 @@ struct CFieldPtrDeleter
...
-    CFieldElement();
+    CFieldElement() = default;
     ~CFieldElement() = default;
-
     explicit CFieldElement(const std::vector<unsigned char>& byteArrayIn);
-    explicit CFieldElement(const uint256& value);
-    void SetByteArray(const std::vector<unsigned char>& byteArrayIn);
+    void SetByteArray(const std::vector<unsigned char>& byteArrayIn) override final;
 
-    CFieldElement(const CFieldElement& rhs) = default;
```
alsala, line 42: We need a custom copy ctor for having the internal data (protected byteVector) copied as well. This also applies to other subclasses when a deep copy is required.

alsala (approved):

i-Alex (approved):

src/sc/sidechaintypes.h
albertog78: Instead of throwing a runtime exception for not having deserialized the expected number of bytes, also considered that both proof and vk will have a variable size, remove the check about the expected size and move it to the concrete class IsValid method.
Remove also the SerializedSize methods from derived classes

albertog78 (changes requested):

ptagl:
This PR is now obsolete.

PR #108, opened 2021-04-26
Support for multiple proving systems

ptagl:
It has been added the possibility to specify (during the sidechain creation) the type of proving system to be used for the verification of certificates and CSWs


qa/rpc-tests/sc_proving_system_selection.py
```diff
@@ -0,0 +1,488 @@
...
+        mark_logs("Node0 generating 1 block", self.nodes, DEBUG_MODE)
+        prev_epoch_block_hash = self.nodes[0].getblockhash(self.nodes[0].getblockcount())
+        blocks.extend(self.nodes[0].generate(1))
+        self.sync_all()
+
+        mark_logs("Verify that sidechain configuration is as expected after connecting the new block", self.nodes, DEBUG_MODE)
+        scinfo0 = self.nodes[0].getscinfo(scid)['items'][0]
+        assert_equal(scinfo0['certificateProvingSystem'], cert_proving_system)
+        assert_equal(scinfo0['cswProvingSystem'], csw_proving_system)
+
```
alsala, line 485: Consider adding a stop/restart phase for checking persistance of the configuration

qa/rpc-tests/sc_proving_system_selection.py
```diff
@@ -0,0 +1,488 @@
...
+    stop_nodes, wait_bitcoinds, \
+    start_nodes, sync_blocks, sync_mempools, connect_nodes_bi, disconnect_nodes, mark_logs, \
+    dump_sc_info_record, get_epoch_data
+from test_framework.mc_test.mc_test import *
+import os
+from decimal import Decimal
+import pprint
+import time
+
+NUMB_OF_NODES = 1
```
alsala, line 18: Some of these constant are not used/useful

src/rpc/blockchain.cpp
```diff
@@ -1203,6 +1208,7 @@ bool FillScRecordFromInfo(const uint256& scId, const CSidechain& info, CSidechai
 
             if (bVerbose)
             {
+                sc.push_back(Pair("certificateProvingSystem", static_cast<uint8_t>(info.fixedParams.certificateProvingSystem)));
```
alsala, line 1211: It would be better to add the prepending 'unconf' suffix as other attributes. The same for the analogous line below.

src/sc/sidechainrpc.cpp
```diff
@@ -440,6 +443,21 @@ bool AddSidechainCreationOutputs(UniValue& sc_crs, CMutableTransaction& rawTx, s
...
         address.SetHex(inputString);
 
+        const UniValue& certProvingSystemValue = find_value(o, "certProvingSystem");
+        if (certProvingSystemValue.isNull() || !certProvingSystemValue.isNum())
+        {
+            error = "Invalid parameter or missing certProvingSystem key";
+            return false;
+        }
+        uint8_t certProvingSystemIndex = certProvingSystemValue.get_int();
+        if (certProvingSystemIndex < static_cast<uint8_t>(Sidechain::ProvingSystemType::CoboundaryMarlin) ||
```
alsala, line 453: Consider using an exact match for all supported enum values instead of a range check. The same applies also for analogous check below.

src/wallet/rpcwallet.cpp
```diff
@@ -887,12 +897,28 @@ UniValue sc_create(const UniValue& params, bool fHelp)
         }
     }
 
-    if (params.size() > 6)
+    if (params.size() > 7)
     {
-        const std::string& inputString = params[6].get_str();
+        uint8_t cswProvingSystemIndex = params[7].get_int();
+
+        if (cswProvingSystemIndex < static_cast<uint8_t>(Sidechain::ProvingSystemType::CoboundaryMarlin) ||
```
alsala, line 904: Consider using an exact match for all supported enum values instead of a range check. The same applies for analogous checks in this file.

alsala (changes requested):

src/coins.cpp
```diff
@@ -814,7 +814,9 @@ bool CCoinsViewCache::UpdateSidechain(const CTransaction& tx, const CBlock& bloc
         scIt->second.sidechain.fixedParams.withdrawalEpochLength = cr.withdrawalEpochLength;
         scIt->second.sidechain.fixedParams.customData = cr.customData;
         scIt->second.sidechain.fixedParams.constant = cr.constant;
+        scIt->second.sidechain.fixedParams.certificateProvingSystem = cr.certificateProvingSystem;
```
i-Alex, line 817: Note: the newly added field of ScCreation output must be verified in
`bool Sidechain::checkTxSemanticValidity(const CTransaction& tx, CValidationState& state)`
method to prevent transactions with invalid ScCreations to be included into mempool and/or state

i-Alex (changes requested):

abi87 (approved):
Approved since I find no other comments than other code reviewers'

src/sc/sidechain.cpp
```diff
@@ -241,12 +251,36 @@ bool Sidechain::checkTxSemanticValidity(const CTransaction& tx, CValidationState
                     CValidationState::Code::INVALID, "sidechain-sc-creation-invalid-constant");
         }
 
-        if (sc.wCeasedVk.is_initialized() && !sc.wCeasedVk.get().IsValid())
+        if (sc.wCeasedVk.is_initialized() )
         {
-            return state.DoS(100,
+            if (!sc.wCeasedVk.get().IsValid())
```
i-Alex, line 256: Does VK "isValid" method logic depend on `ProvingSystemType`?
For example, Vk was created for Darlin system, so we expect the key to be valid for Darlin and not valid for CoboundaryMarlin. Or is it valid for both?
Does it make any sense? @DanieleDiBenedetto

i-Alex (changes requested):

i-Alex (changes requested):

i-Alex (approved):

ptagl:
These changes have already been integrated into `sidechains_integration_step4`.

PR #117, opened 2021-05-21
Sidechains integration step temp PR

albertog78:


src/sc/proofverifier.cpp
```diff
@@ -46,129 +44,209 @@ void CScProofVerifier::LoadDataForCertVerification(const CCoinsViewCache& view,
...
     }
 
-    certData.quality = scCert.quality; //Currently quality not yet accounted for in proof verifier
-    if (sidechain.fixedParams.constant.is_initialized())
-        certData.constant = sidechain.fixedParams.constant.get();
-    else
-        certData.constant = CFieldElement{};
+    for (auto entry: scCert.vFieldElementCertificateField)
+    {
+        CFieldElement fe{entry.getVRawData()};
```
i-Alex, line 49: Per my understanding `entry.getVRawData()` returns raw bytes of custom fields, that are not exactly a field element bytes, but can be less. So `CFieldElement fe{entry.getVRawData()}` leads to the invalid FieldElement and moreover asserts inside `CFieldElement` constructor.
I think we need to use `entry.GetFieldElement` method.

src/sc/proofverifier.cpp
```diff
@@ -46,129 +44,209 @@ void CScProofVerifier::LoadDataForCertVerification(const CCoinsViewCache& view,
...
+        certData.vCustomFields.push_back(fe);
+    }
+    for (auto entry: scCert.vBitVectorCertificateField)
+    {
+        CFieldElement fe{entry.getVRawData()};
```
i-Alex, line 54: Same here instead of passing `entry.getVRawData()` we should use `GetFieldElement` method that will return the merkle root of the compressed bitvector.

i-Alex (changes requested):

src/sc/proofverifier.cpp
alsala: It is true that the CFieldElement obj stores internally the raw representation of data, but when it is used for input to the prover, its deserialized version is used via CFieldElement()::GetFieldElement() method.

src/sc/proofverifier.cpp
alsala: Same as above

src/sc/proofverifier.cpp
i-Alex: I think, yes.
The important thing is that CFieldElement stores exactly 32 bytes of raw data.
From other side `FieldElementCertificateField` stores up to 32 bytes of raw data.

For example, it can store only 4 bytes representing some int value.
So, `FieldElementCertificateField::getVRawData()` will return vector with the length 4.
The expected behavior is to return the full field element 32 bytes and then use them to create a `CFieldElement` obj.

src/init.cpp
```diff
@@ -1227,6 +1239,11 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     // Initialize Zcash circuit parameters
     ZC_LoadParams();
 
+    // check type sizes in crypto lib are as expected and assert() in case of failure
+    CZendooCctpLibraryChecker::CheckTypeSizes();
+
+    Sidechain::LoadCumulativeProofsParameters();
```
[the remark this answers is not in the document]
ptagl: Fixed.

src/primitives/certificate.cpp
```diff
@@ -315,7 +314,10 @@ CScCertificate::MakeShared() const {
 
 CFieldElement CScCertificate::GetDataHash() const
 {
-    std::vector<unsigned char> tmp(this->GetHash().begin(), this->GetHash().end());
+    // in the final implementation this must not create an obj but should return a reference to a mem-only
+    // data member, similarly to what GetHash() is doing. This is for performances but expecially for avoiding 
+    // having temporary objs and playing with their internal data buffers. 
+    std::vector<unsigned char> tmp(this->GetHash().begin(), this->GetHash().end() - 2);
```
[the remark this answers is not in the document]
ptagl: Fixed.

src/sc/asyncproofverifier.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+        cswData.certDataHash = csw.actCertDataHash;
+        cswData.ceasingCumScTxCommTree = csw.ceasingCumScTxCommTree;
+
+        cswData.cswProof = csw.scProof;
+
+        CSidechain sidechain;
+        assert(view.GetSidechain(csw.scId, sidechain) && "Unknown sidechain at scTx proof verification stage");
+
+        if (sidechain.fixedParams.wCeasedVk.is_initialized())
```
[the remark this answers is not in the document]
ptagl: Fixed.

src/sc/proofverifier.cpp
ptagl: Fixed by implementing a new method called getExtendedRawData() that always returns 32 bytes (eventually padding). The other approach of using GetFieldElement was not implementable since there are cases in which the fixed params are not immediately available (e.g. when updating the cumulative commitment tree).

In the future, we may consider storing the custom fields as 32 bytes field elements even though the real data is smaller; would it be ok or do we need to save as much memory as we can on mainchain side? @albertog78 

In any case this activity should be performed together with the refactoring of the part related to the conversion of data for usage with the CCTP Lib interfaces (e.g. to avoid using `wrappedFieldPtr` in the `CScProofVerifier` and `CScAsyncProofVerifier`).

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -4,130 +4,486 @@
...
+        vbt_list.push_back(x);
+    }
+
+    if (!vbt_list.empty())
+        bt_list = (const backward_transfer_t*)vbt_list.data();
+
+    size_t bt_list_len = vbt_list.size();
+
+    int custom_fields_len = cert.vFieldElementCertificateField.size(); 
+    std::unique_ptr<const field_t*[]> custom_fields(new const field_t*[custom_fields_len]);
```
i-Alex, line 246: `custom_fields` should keep bitvectors as well.

PR #122, opened 2021-06-03
Ported fix related to p2p mempool msg from PR107 into sidechains_inte…

alsala:
…gration_step4 branch; a variant of the UT logic with lower impact on code base has been imported as well

### DIFF
--- a/src/gtest/test_mempool.cpp
+++ b/src/gtest/test_mempool.cpp
@@ -372,7 +372,7 @@ public:
 
     std::set<CInv> pushedInvList;
 
-    void PushInvs(const char* pszCommand, const std::vector<CInv>& invVec)
+    void PushInvs(const char* pszCommand, const std::vector<CInv>& invVec) override
     {
         for (auto const & inv : invVec)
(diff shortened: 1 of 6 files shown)


i-Alex (approved):

src/gtest/test_mempool.cpp
```diff
@@ -349,3 +350,62 @@ TEST(Mempool, SproutNegativeVersionTx) {
...
+    }
+
+    CNodeExt():
+        CNode(INVALID_SOCKET, CAddress(ip(0xa0b0c002)), "", true)
+    {
+    }
+
+    std::set<CInv> pushedInvList;
+
+    void PushInvs(const char* pszCommand, const std::vector<CInv>& invVec)
```
ptagl, line 375: To avoid possible compilation warnings and to improve readability, I would add the `override` specifier to this method.

```suggestion
    void PushInvs(const char* pszCommand, const std::vector<CInv>& invVec) override
```

ptagl (approved):

Issue #95, opened 2021-03-05
Wallet CreateTransaction CSWs full support


i-Alex:
Current implementation of `bool CWallet::CreateTransaction` has several issues with CSWs:
1) If the initial transaction (like the one passed via `fundrawtransaction`) contains CSWs which total amount already covers total outputs amount (vouts and ccouts), `CreateTransaction` will try to fund a negative value of `totalOutputValue`
2) If the initial transaction contains CSWs and the `sign` argument of the method is `true`, the fund inputs message to sign will not include CSWs data, so the resulting signatures will be invalid.

ptagl:
This issue has been fixed with PR #126.

PR #126, opened 2021-06-11
As/issue 95

alsala:
Fix and py test for issue #95: added csw input to CreateTransaction wallet function in order to handle properly csw in the fundrawtransaction cmd


src/wallet/wallet.cpp
```diff
@@ -3489,6 +3492,9 @@ bool CWallet::CreateTransaction(
                     }
                 }
 
+                // add csw input
```
i-Alex, line 3495: See the code above related to the `mempooltxinputlimit` check. We should consider CSW inputs as well.

i-Alex (changes requested):

i-Alex (approved):

ptagl (approved):

PR #128, opened 2021-06-15
Sidechains integration step4 cctp deser optimization

alsala:


src/main.cpp
```diff
@@ -2966,6 +2968,9 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
 
     CCheckQueueControl<CScriptCheck> control(fExpensiveChecks && nScriptCheckThreads ? &scriptcheckqueue : NULL);
 
+    int64_t deltaPreProcTime = GetTimeMicros() - nTime0;
+    LogPrint("bench", "    - block preproc: %.2fms\n", 0.001 * deltaPreProcTime);
```
ptagl, line 2972: Since time durations are often converted to milliseconds, I suggest using the function `GetTimeMillis()` and removing the conversion.

src/sc/sidechaintypes.cpp
```diff
@@ -215,9 +242,21 @@ wrappedScProofPtr CScProof::GetProofPtr() const
...
+
+    wrappedScProofPtr res{nullptr};
+    {
+        LOCK(cs);
+        LogPrintf("%s():%d - obj[%p]: checks %s in deserialization call\n", __func__, __LINE__, this, firstDeserializeCall?"ON":"OFF");
+ 
+        res = wrappedScProofPtr{zendoo_deserialize_sc_proof(&result, firstDeserializeCall, &code), theProofPtrDeleter};
+ 
+        // we do the semantic checks only the very first call
+        if (firstDeserializeCall)
```
ptagl, line 256: Instead of using the boolean flag `firstDeserializeCall`, we could use the "wrapped pointer" as a class member and distinguish between the first and next calls by checking if the smart pointer has been initialized or not.

Such an improvement could be beneficial for two reasons:

1. In case `GetProofPtr()` is called more than once, we directly return the smart pointer instead of generating a new one (even though with the current implementation it takes a short time since the expensive deserialization checks are disabled);
2. It would let us simplify the code of some functions, for instance the proof verification functions in `proofverifier.cpp` and `asyncproofverifier.cpp`, where we currently have to create temporary smart pointer "wrappers".

@albertog78 What do you think about that?

ptagl (approved):

src/sc/sidechaintypes.cpp
i-Alex: Good improvement, as for me.
Another approach is to review the interface in general.
It seems, that `check semantics` flag must be set to `true` only for `isValid` method. In other places we may relay that `isValid` was called before. We should consider the duplication of `isValid` execution calls in different place for the same object, so may cache privately the result of it.

alsala: As discussed with ptagl and i-Alex the first approach has been implemented

i-Alex (approved):

PR #129, opened 2021-06-16
As/sc proof batch ver zend options

alsala:
Added zend options for tuning size and delay params for proof batch verification queue

### DIFF
--- a/src/sc/asyncproofverifier.cpp
+++ b/src/sc/asyncproofverifier.cpp
@@ -7,6 +7,8 @@
 #include "primitives/certificate.h"
 #include "sc/proofverifier.h"
 
+const uint32_t CScAsyncProofVerifier::BATCH_VERIFICATION_MAX_DELAY = 5000;   /**< The maximum delay in milliseconds between batch verification requests */
+const uint32_t CScAsyncProofVerifier::BATCH_VERIFICATION_MAX_SIZE = 10;      /**< The threshold size of the proof queue that triggers a call to the batch verification. */
 
 #ifdef BITCOIN_TX
 void CScProofVerifier::LoadDataForCertVerification(const CCoinsViewCache& view, const CScCertificate& scCert, CNode* pfrom) {return;}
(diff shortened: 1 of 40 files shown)


src/sc/asyncproofverifier.cpp
```diff
@@ -72,6 +74,15 @@ void CScAsyncProofVerifier::RunPeriodicVerification()
...
 
+    int32_t delay = GetArg("-scproofverificationdelay", BATCH_VERIFICATION_MAX_DELAY);
+    assert(delay >= 0 && "scproofverificationdelay must be non negative");
+    uint32_t batchVerificationMaxDelay = static_cast<uint32_t>(delay);
+
+    int32_t size = GetArg("-scproofqueuesize", BATCH_VERIFICATION_MAX_SIZE);
+    assert(size >= 0 && "scproofqueuesize must be non negative");
+    uint32_t batchVerificationMaxSize = static_cast<uint32_t>(size);
+
+
```
ptagl, line 85: I wonder if we should restrict these two parameters to work in REGTEST mode only.

ptagl (changes requested):

src/sc/asyncproofverifier.cpp
alsala: I think they could be useful also for other network modes, and anyway a wrong usage should not harm other nodes. Probably we could add a range check for avoid absurd settings

src/sc/asyncproofverifier.h
```diff
@@ -157,7 +157,9 @@ class TEST_FRIEND_CScAsyncProofVerifier
      */
     uint32_t GetMaxBatchVerifyDelay()
     {
-        return CScAsyncProofVerifier::GetInstance().BATCH_VERIFICATION_MAX_DELAY;
+        int32_t delay = GetArg("-scproofverificationdelay", CScAsyncProofVerifier::BATCH_VERIFICATION_MAX_DELAY);
+        assert(delay >= 0 && "scproofverificationdelay must be non negative");
+        return static_cast<uint32_t>(delay);
```
ptagl, line 162: In order to avoid code duplication, we could define a private getter method inside `CScAsyncProofVerifier` and use it both here and in `CScAsyncProofVerifier::RunPeriodicVerification()`.

ptagl (changes requested):

src/sc/asyncproofverifier.h
alsala: Done, I also turned assert() into condition check

ptagl (approved):

PR #146, opened 2021-07-05
Sidechains integration step4 code review

ptagl:
This PR shows the differences between the `sidechains_integration_step4_code_review` and  `sidechains_integration_code_review`. The latter is the union between `sidechains_integration_step2` and two backports:

- `backport_zen_2.0.22_sidechains_integration_step4`
- `backport_zen_2.0.23_sidechains_integration_step4`


src/script/interpreter.cpp
```diff
@@ -1432,38 +1432,7 @@ bool CertificateSignatureChecker::CheckSig(const vector<unsigned char>& vchSigIn
...
-    CBlockIndex* pblockindex = (*chain)[nHeight];
-    uint256 blockHash = pblockindex->GetBlockHash();
-    std::vector<unsigned char> vchBlockHash(blockHash.begin(), blockHash.end());
-
-    if (vchBlockHash.empty() || vchCompareTo.empty()) {
-        return false;
-    }
-
-    return (vchCompareTo == vchBlockHash);
+    return CheckReplayProtectionData(chain, nHeight, vchCompareTo);
```
i-Alex, line 1435: Was not able to find the function definition in the current PR. I guess that it was defined in the base branch.

i-Alex (approved):

alsala (approved):

albertog78 (approved):

src/script/interpreter.cpp
ptagl: Yes, it was already defined in the same file at [line 1352](https://github.com/HorizenOfficial/zend_oo/blob/497d40ab503854827efbb8021e457ba43a372ad2/src/script/interpreter.cpp#L1352); it comes from a backported patch of Zend.

PR #147, opened 2021-07-06
Sidechains integration step4 low prio thr pause

alsala:
Added a RAII guard for pausing rust low prio threads when connecting a block

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2920,6 +2920,16 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         }
     }
 
+    bool pauseLowPrioZendooThread = (
+        fExpensiveChecks &&
+        fScRelatedChecks == flagScRelatedChecks::ON &&
+        fScProofVerification == flagScProofVerification::ON &&
+        SidechainTxsCommitmentBuilder::getEmptyCommitment() != block.hashScTxsCommitment // no sc related tx/certs 
+    );
+       
+    // if necessary pause rust low priority threads in order to speed up times
+    CZendooLowPrioThreadGuard lowPrioThreadGuard(pauseLowPrioZendooThread);
+     
     auto verifier = libzcash::ProofVerifier::Strict();
     auto disabledVerifier = libzcash::ProofVerifier::Disabled();
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2926,6 +2926,10 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         fScProofVerification == flagScProofVerification::ON &&
         SidechainTxsCommitmentBuilder::getEmptyCommitment() != block.hashScTxsCommitment // no sc related tx/certs 
     );
+
+    // if necessary pause rust low priority threads in order to speed up times
+    // Note: it works even if the same code was executed for the high priority proof verifier
+    CZendooLowPrioThreadGuard lowPrioThreadGuard(pauseLowPrioZendooThread);
      
     auto verifier = libzcash::ProofVerifier::Strict();
     auto disabledVerifier = libzcash::ProofVerifier::Disabled();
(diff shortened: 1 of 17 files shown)


src/sc/proofverifier.cpp
```diff
@@ -77,7 +77,7 @@ bool CScProofVerifier::BatchVerify() const
     }
 
     CctpErrorCode code;
-    ZendooBatchProofVerifier batchVerifier;
+    ZendooBatchProofVerifier batchVerifier(true);
```
i-Alex, line 80: what is the meaning of constructor arg equals `true`?
Do we need to change the creation policy in other places, like in `asyncproofverifier::BatchVerify`?

alsala: The paramenter in the ctor is a boolean telling mc-crypto lib if the rust verifier executing thread will be a high priority one. We can probably add a comment here or a static const var with a speaking name.

i-Alex: Ok, I got it. Would be nice to add comment or const.

DanieleDiBenedetto (approved):

src/sc/proofverifier.cpp
alsala: Done

i-Alex (approved):

ptagl (approved):

src/sc/proofverifier.cpp
albertog78: It seems that both the CscAsyncProofVerifier and CScProofVerifier instantiate the threads with high priority

albertog78 (changes requested):

src/main.cpp
```diff
@@ -2920,6 +2920,16 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
     }
 
+    bool pauseLowPrioZendooThread = (
+        fExpensiveChecks &&
+        fScRelatedChecks == flagScRelatedChecks::ON &&
+        fScProofVerification == flagScProofVerification::ON &&
+        SidechainTxsCommitmentBuilder::getEmptyCommitment() != block.hashScTxsCommitment // no sc related tx/certs 
+    );
+       
+    // if necessary pause rust low priority threads in order to speed up times
```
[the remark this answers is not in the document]
i-Alex: Actually we don't need `CZendooLowPrioThreadGuard` at all.
Instead we should create `ProofVerifier` with one more argument `prioritize` same as in mc-cryptolib.
When the prioritized `batchVerified` is triggered, it will automatically pause then resume the low priority threads.

src/sc/proofverifier.cpp
i-Alex: Right. We must add one more constructor argument to `CScProofVerifier` -> `prioritize` and set it `true` for connect block, `false` for mempool verification.
Moreover, `CScAsyncProofVerifier` should have `prioritize = false` as it is used only for mempool txs/certs.
@albertog78 priority test is present in mc-cryptolib `zendoo_batch_verifier_multiple_threads_with_priority`

i-Alex (changes requested):

i-Alex (approved):

src/sc/proofverifier.cpp
```diff
@@ -231,15 +231,15 @@ bool CScProofVerifier::BatchVerifyInternal(std::map</* Cert or Tx hash */ uint25
         }
     }
 
-    CctpErrorCode code;
-    ZendooBatchProofVerifier batchVerifier;
+    // The paramenter in the ctor is a boolean telling mc-crypto lib if the rust verifier executing thread
```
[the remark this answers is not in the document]
albertog78: Yes seems a bug. @alsala @i-Alex

albertog78 (changes requested):

src/sc/proofverifier.cpp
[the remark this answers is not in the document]
alsala: Agree, it should be outside the for loop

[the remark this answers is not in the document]
i-Alex: done

albertog78 (approved):

PR #165, opened 2021-07-30
As/zoo 004

alsala:
Added a constant in chain params for getting the number of recent SC active data certificates whose scFee values are to be include in a list for checking ft/mbtr sc fees and deciding if a tx should be removed from mempool when a certificate is included in a block just connected; the check is done agaist the minimum scFee in such list.


src/sc/sidechain.cpp
```diff
@@ -503,4 +506,126 @@ bool CSidechain::GetCeasingCumTreeHash(CFieldElement& ceasedBlockCum) const
...
+        sizeOfScFeesContainers = numBlocks / epochLength;
+        if (sizeOfScFeesContainers == 0 || (numBlocks % epochLength != 0) )
+        {
+            sizeOfScFeesContainers++;
+        }
+        LogPrint("sc", "%s():%d - sizeOfScFeesContainers set to %d\n", __func__, __LINE__, sizeOfScFeesContainers);
+
+        assert(sizeOfScFeesContainers > 0);
+        assert(scFees.empty());
+        scFees.resize(sizeOfScFeesContainers);
```
ptagl, line 557: Should we use the FT fee and MBTR fee values to initialize the `scFees` list instead of 0? I guess that by using 0 it could be possible to include FT or MBTR with 0 fees in a block during the first N epochs, since we have a different check for transactions in `AcceptTxToMemoryPool()` and in `ConnectBlock()`.

I see two possible solutions:

- initializing the scFees list with the N values equal to the fees declared during sidechain creation

- initializing the scFees list with one single value (equal to the fees declared during sidechain creation) and gradually increasing the size of the list until we reach the epoch number N (and the list is full)

ptagl (changes requested):

src/sc/sidechain.cpp
alsala: Agree; I chose the first option. I also moved the InitScFees() method call into sc creation handling

src/sc/sidechain.cpp
```diff
@@ -503,4 +506,128 @@ bool CSidechain::GetCeasingCumTreeHash(CFieldElement& ceasedBlockCum) const
...
+        assert(scFees.empty());
+        Sidechain::ScFeeData defaultData(
+            lastTopQualityCertView.forwardTransferScFee, lastTopQualityCertView.mainchainBackwardTransferRequestScFee);
+        scFees.resize(sizeOfScFeesContainers, defaultData);
+    }
+}
+
+void CSidechain::UpdateScFees(const CScCertificateView& certView)
+{
+    InitScFees();
```
i-Alex, line 565: Why we need to call `InitScFees()`? It should be initialized only by the ScCreation output.

src/sc/sidechain.cpp
```diff
@@ -503,4 +506,128 @@ bool CSidechain::GetCeasingCumTreeHash(CFieldElement& ceasedBlockCum) const
...
+        std::cout << "[" << std::setw(2) << entry.forwardTxScFee
+                  << "/" << std::setw(2) << entry.mbtrTxScFee << "]";
+    }
+    std::cout << std::endl;
+}
+
+CAmount CSidechain::GetMinFtScFee() const
+{
+    CAmount minScFee = MAX_MONEY;
+    if (scFees.empty())
```
i-Alex, line 600: How scFees can be empty? It should be equals to the SC Creation output value at least.

src/sc/sidechain.cpp
```diff
@@ -503,4 +506,128 @@ bool CSidechain::GetCeasingCumTreeHash(CFieldElement& ceasedBlockCum) const
...
+    return minScFee;
+}
+
+CAmount CSidechain::GetMinMbtrScFee() const
+{
+    CAmount minScFee = MAX_MONEY;
+    if (scFees.empty())
+    {
+        LogPrint("sc", "%s():%d - list empty, returning creation value: %lld\n", __func__, __LINE__, minScFee);
+        minScFee = pastEpochTopQualityCertView.mainchainBackwardTransferRequestScFee;
```
i-Alex, line 621: Same here.

src/sc/sidechain.h
```diff
@@ -103,6 +104,14 @@ class CSidechain {
     // value = the immature amount
     std::map<int, CAmount> mImmatureAmounts;
 
+    // the last ftScFee and mbtrScFee values, as set by the active certificates
+    std::list<Sidechain::ScFeeData> scFees;
+
+    // memory only
+    int sizeOfScFeesContainers;
```
i-Alex, line 111: As I understood it is the `maxSize...`. Let's rename

src/sc/sidechain.h
```diff
@@ -131,22 +140,29 @@ class CSidechain {
         READWRITE(balance);
         READWRITE(fixedParams);
         READWRITE(mImmatureAmounts);
+        READWRITE(scFees);
+        if (ser_action.ForRead())
+        {
+            if (!scFees.empty())
+                sizeOfScFeesContainers = scFees.size();
```
i-Alex, line 147: `sizeOfScFeesContainers` should depend on the `getNumBlocksForScFeeCheck` and epoch length, not on `scFees.size`.
For example, `sizeOfScFeesContainers` was initialized  = 10, but the `scFees` consists of 3 elements (we had just a few epochs). After the node restart -> the value `sizeOfScFeesContainers == 3` instead of 10.

i-Alex: We may even store `sizeOfScFeesContainers` value as well to avoid recalculation of that value, but can lead to the problem when we set the `-blocksforscfeecheck` flag on regtest to decrease the `sizeOfScFeesContainers`.

src/sc/sidechain.cpp
```diff
@@ -503,4 +506,128 @@ bool CSidechain::GetCeasingCumTreeHash(CFieldElement& ceasedBlockCum) const
...
+
+void CSidechain::UpdateScFees(const CScCertificateView& certView)
+{
+    InitScFees();
+
+    CAmount ftScFee   = certView.forwardTransferScFee;
+    CAmount mbtrScFee = certView.mainchainBackwardTransferRequestScFee;
+
+    // check if we are past the buffer size
+    if (scFees.size() >= sizeOfScFeesContainers)
```
i-Alex, line 571: Consider the case with `sizeOfScFeesContainers` decrement if in regtest node was restarted with the flag `-blocksforscfeecheck` less than before, it may need to remove more than one element.

i-Alex (changes requested):

src/sc/sidechain.cpp
```diff
@@ -503,4 +506,126 @@ bool CSidechain::GetCeasingCumTreeHash(CFieldElement& ceasedBlockCum) const
     return true;
 }
 
+#if 0
+static int getInitNumBlocksForScFeeCheck()
+#else
```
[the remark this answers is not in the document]
alsala: Done

src/sc/sidechain.cpp
```diff
@@ -503,4 +506,126 @@ bool CSidechain::GetCeasingCumTreeHash(CFieldElement& ceasedBlockCum) const
...
+#endif
+
+#if 0
+int CSidechain::getNumBlocksForScFeeCheck()
+{
+    // gets constructed just one time
+    static int retVal( getInitNumBlocksForScFeeCheck() );
+    return retVal;
+}
+#endif
```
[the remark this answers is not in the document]
alsala: Removed

qa/rpc-tests/sc_stale_ft_and_mbtr.py
```diff
@@ -0,0 +1,532 @@
...
+# Copyright (c) 2018 The Zencash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, assert_true, assert_false, initialize_chain_clean, \
+    stop_nodes, wait_bitcoinds, \
+    start_nodes, sync_blocks, sync_mempools, connect_nodes_bi, disconnect_nodes, mark_logs, \
+    dump_sc_info_record, get_epoch_data, swap_bytes, advance_epoch
+from test_framework.mc_test.mc_test import *
```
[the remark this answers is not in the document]
alsala: Done

src/sc/sidechain.cpp
alsala: After further discussion, the second approach has been choosen.

src/sc/sidechain.cpp
alsala: Removed. As a consequence, some UT has been modified explicitly adding InitScFees().

src/sc/sidechain.cpp
alsala: Modified into assert() condition checking, also because there is no default behaviour in that case

src/sc/sidechain.cpp
alsala: Done

src/sc/sidechain.h
alsala: Done

src/sc/sidechain.h
alsala: Moved the logic for computing such value and called also in this section.

src/sc/sidechain.cpp
alsala: Ok, done

ptagl (approved):

i-Alex (approved):

albertog78 (approved):

PR #173, opened 2021-08-17
Fix for certificates handling in GetBlockTemplate

ptagl:
The GetBlockTemplate function didn't take into account certificates added to the mempool, but only transactions. For this reason, if the mempool got updated with just new certificates, the GetBlockTemplate function returned an "empty" block.


ptagl:
This PR also fixes https://github.com/HorizenOfficial/zen/issues/349.

cronicc (changes requested):
Add qa/rpc-tests/sc_cert_getblocktemplate.py to qa/pull-tester/rpc-tests.sh

cronicc (approved):

PR #181, opened 2021-09-03
Issue#177 + WalletTx::GetAmounts(...)

i-Alex:


src/wallet/rpcwallet.cpp
```diff
@@ -2886,6 +2879,11 @@ UniValue listtransactions(const UniValue& params, bool fHelp)
         }
     }
 
+    bool includeImmatureBTs = false;
+    if(params.size() > 5)
+        if(params[5].getBool())
```
alsala, line 2884: A minor change: it might be useful add isBool() check, otherwise a user can set '1' meaning 'True' and the cmd executes 'False'.
Or better: use get_bool() which throws an exception directly

alsala (changes requested):

src/wallet/rpcwallet.cpp
i-Alex: done

PR #184, opened 2021-09-07
Added PROOF_PLUS_VK_SIZE constant and check it for certs and csws. Bug fix in proof/vk isValid().

DanieleDiBenedetto:
Still need to address:

- Add a python test that checks that IsScTxApplicableToState and IsCertApplicableToState fail when proof + vk size exceeds the new limit just introduced. To do this we need to generate big proofs and vks in the python tests. It's enough to call the already existing functions using a small segment size and a big number of constraints;
- In zendoo-mc-cryptolib we need to return error whenever we deserialize succesfully variable sized data but there are still bytes to read.

### DIFF
--- a/src/gtest/test_libzendoo.cpp
+++ b/src/gtest/test_libzendoo.cpp
@@ -1928,3 +1928,51 @@ TEST(CctpLibrary, TestVectorsValidity)
     EXPECT_TRUE(p4.IsValid());
 }
 
+//TODO: Maybe it's not the correct place for this test
+TEST(CctpLibrary, TestInvalidProofVkWhenOversized)
+{
+    // Oversized Proof
+    std::vector<unsigned char> OVERSIZED_CERT_DARLIN_PROOF = SAMPLE_CERT_DARLIN_PROOF;
+    OVERSIZED_CERT_DARLIN_PROOF.resize(OVERSIZED_CERT_DARLIN_PROOF.size() + Sidechain::MAX_SC_PROOF_SIZE_IN_BYTES, 0xAB);
+    EXPECT_DEATH(CScProof{OVERSIZED_CERT_DARLIN_PROOF}, "");
+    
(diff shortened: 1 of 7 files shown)


src/sc/sidechaintypes.cpp
```diff
@@ -304,6 +304,14 @@ wrappedScProofPtr CScProof::GetProofPtr() const
     std::lock_guard<std::mutex> lk(_mutex);
     if (proofData == nullptr)
     {
+        if (byteVector.size() > MaxByteSize())
```
i-Alex, line 307: As I saw, we have an assert in CScProof constructor and `SetByteArray` method:
`assert(byteArrayIn.size() <= this->MaxByteSize());`
Do we need additional check here?

i-Alex (changes requested):

ptagl (approved):

src/gtest/test_libzendoo.cpp
```diff
@@ -1928,3 +1928,51 @@ TEST(CctpLibrary, TestVectorsValidity)
...
+    INVALID_CERT_DARLIN_VK.push_back(0xAB);
+    EXPECT_NO_FATAL_FAILURE(CScVKey{INVALID_CERT_DARLIN_VK});
+
+    stream << INVALID_CERT_DARLIN_VK;
+    CScVKey vkInvalid;
+    stream >> vkInvalid;
+    //TODO: Currently fails. We need to modify mc-cryptolib in order to return error even the proof
+    //      is deserialized correctly, but there are still bytes to read.
+    //EXPECT_FALSE(vkInvalid.IsValid());
+}
```
ptagl, line 1978: It might be useful to add another test to check that we don't have the same issue with the BitVector.

ptagl (changes requested):

albertog78 (approved):

i-Alex (approved):

ptagl (approved):

PR #190, opened 2021-09-13
RPC changes: pubkeyhash replaced with taddr

i-Alex:
FT, MBTR, CSW, Certs: direct usage of pubkeyhash removed. Now transparent addresses are used in RPC commands and JSON.

### DIFF
--- a/src/zendoo/mcTestCall.cpp
+++ b/src/zendoo/mcTestCall.cpp
@@ -6,6 +6,54 @@
 #include <cassert>
 #include <string>
 
+// TODO: remove as soon we will be able to link base56.cpp
+static const char* pszBase58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+bool DecodeBase58(const char* psz, std::vector<unsigned char>& vch)
+{
+    // Skip leading spaces.
+    while (*psz && isspace(*psz))
+        psz++;
+    // Skip and count leading '1's.
+    int zeroes = 0;
+    while (*psz == '1') {
+        zeroes++;
+        psz++;
+    }
+    // Allocate enough space in big-endian base256 representation.
+    std::vector<unsigned char> b256(strlen(psz) * 733 / 1000 + 1); // log(58) / log(256), rounded up.
+    // Process the characters.
+    while (*psz && !isspace(*psz)) {
+        // Decode base58 character
+        const char* ch = strchr(pszBase58, *psz);
+        if (ch == NULL)
+            return false;
+        // Apply "b256 = b256 * 58 + ch".
+        int carry = ch - pszBase58;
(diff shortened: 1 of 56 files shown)


src/zendoo/mcTestCall.cpp
```diff
@@ -141,10 +189,17 @@ void create_verify_test_cert_proof(std::string ps_type_raw, std::string cert_typ
         for(int i = 0; i < bt_list_length; i ++){
             backward_transfer_t bt;
 
-            assert(IsHex(argv[arg]));
+            std::vector<unsigned char> vchData;
+            assert(DecodeBase58(argv[arg++], vchData));
```
cronicc, line 193: Do we need [DecodeBase58Check](https://github.com/HorizenOfficial/zend_oo/blob/4eb908f4453ae9a1131676f4dc5cf81f4f46087a/src/base58.cpp#L124-L139) in addition to DecodeBase58? If the inputs are normal zen addresses they'd be not just Base58 encoded but they'd include a checksum.

i-Alex: I think, for test tool is enough for now.
As you can mention `mcTestCall` duplicated the code of `DecodeBase58`. That's temporary until we solve the problem with the dep libs linkage, for some reason it raise errors. Then we will be able to use CBitcoinAddress with all the checks.

alsala (approved):

src/zendoo/mcTestCall.cpp
```diff
@@ -360,11 +415,20 @@ void create_verify_test_csw_proof(std::string ps_type_raw, std::string csw_type_
...
-    // Parse mc_pk_hash
-    assert(IsHex(argv[arg]));
+    // Parse mc_address
+    // Extract pubKeyHash from the address
+    std::vector<unsigned char> vchData;
+    assert(DecodeBase58(argv[arg++], vchData));
     uint160 mc_pk_hash_vec;
-    mc_pk_hash_vec.SetHex(argv[arg++]);
-    assert(mc_pk_hash_vec.size() == 20);
+    memcpy(&mc_pk_hash_vec, &vchData[2], 20);
```
VolodymyrBabaiev, line 423: Consider to change name mc_pk_hash_vec  to pk_dest as it implemented in create_verify_test_cert_proof function. The same for mc_pk_hash variable below.

i-Alex: done

VolodymyrBabaiev (approved):

albertog78 (approved):

cronicc:
@i-Alex please resolve conflicts by
```
git checkout testnet_staging
git pull
git checkout pubkeyhash_removal
git rebase testnet_staging
```
Then resolve conflicts similar to a merge, git status, edit, git add, git rebase --continue.

i-Alex:
Done

PR #191, opened 2021-09-15
Added version in gettransaction and CTP fields only if it's a sidecha…

MarcoOl94:
1. Added field "version" inside gettransaction
2. The CTP fields are included only if the transaction has version -4


src/wallet/rpcwallet.cpp
```diff
@@ -3460,6 +3532,7 @@ UniValue gettransaction(const UniValue& params, bool fHelp)
         nFee = -(wtx.getTxBase()->GetFeeAmount(nDebit) + cswInTotAmount);
     }
 
+    entry.pushKV("version", wtx.getTxBase()->nVersion);
```
i-Alex, line 3535: Put this line inside `WalletTxToJSON` method.

i-Alex (changes requested):

albertog78 (approved):

src/wallet/rpcwallet.cpp
albertog78: check where is WalletTxToJSON is called

Issue #155, opened 2021-07-26
sc_create transaction doesn't have 1sat/byte tx fee


cronicc:
The expected behavior for transactions that calculate the transaction fee based on transaction size is to use a rate of 1sat/byte. This is the case for any transparent transactions created using e.g. `sendtoaddress` or `sendmany`.

In case of `sc_create` I encountered a transaction fee rate of ~1.289sat/byte.

Is setting a slightly higher fee rate the intended behavior or do we have an issue with the size estimation/calculation of sc_create transactions?

Note: I haven't checked other SC related MC transactions for fee behavior yet.

Sidechain creation tx, single 1.0001 ZEN vin, 1.00009023 ZEN FT, exact spend with no change, fee of 977sat. Transaction size was 758 bytes:
```
{
  "txid": "ea3dd4c9e08454451760593a947978c4a2a59bfb441cc4e94bc88c1d1d4500af",
  "version": -4,
  "locktime": 876735,
  "vin": [
    {
      "txid": "9179cd4c77633f316dff9e4bc93730486c94a86c6919fd04805b464e01b1403e",
      "vout": 213,
      "scriptSig": {
        "asm": "3045022100c84a0e3f298b038a1719b98e8a2fd355d4db310966ac7f80da5c412da1c2b4a502204911c0a983f51570dab31f05da2b57dea5e7ab7e906a65bf7ec6f75b5b53414301 03ab0a9262e7f61fc12a95ce5424ea18bcd3f3a7dcdc4d8cff7816143391471d01",
        "hex": "483045022100c84a0e3f298b038a1719b98e8a2fd355d4db310966ac7f80da5c412da1c2b4a502204911c0a983f51570dab31f05da2b57dea5e7ab7e906a65bf7ec6f75b5b534143012103ab0a9262e7f61fc12a95ce5424ea18bcd3f3a7dcdc4d8cff7816143391471d01"
      },
      "sequence": 4294967294
    }
  ],
  "vcsw_ccin": [
  ],
  "vout": [
  ],
  "vsc_ccout": [
    {
      "scid": "2441cde2fdef217668ae19992486d10aa751cb666359cce6a8028616ef44dea6",
      "n": 0,
      "withdrawal epoch length": 100,
      "value": 1.00009023,
      "address": "f59bb3af7d09446c490a000be2a9ff4dac4f8b8c931a153550f47b8587ea9c52",
      "certProvingSystem": "CoboundaryMarlin",
[...]

PR #207, opened 2021-10-12
Automatic fee computation for SC related rpc commands

alsala:
Solves issue #155. Added automatic fee computation if fee is not explicitly set by the user in rpc cmds related to SC. The fee rate is by default 1 Sat per byte as in bitcoin rpc cmds implementation.


qa/rpc-tests/sc_rpc_cmds_fee_handling.py
```diff
@@ -0,0 +1,293 @@
...
+        Test the automatic fee computation for SC related transactions and certificates
+        '''
+
+        def get_fee_rate(size, fee):
+            return ((fee*COIN)/size)
+        
+        def isclose(d1, d2, tolerance=Decimal('0.005')):
+            dec1 = Decimal(d1)
+            dec2 = Decimal(d2)
+            return abs(d1-d2) <= tolerance
```
VolodymyrBabaiev, line 65: In function abs used non-converted values of d1 and d2 instead of dec1 and dec2.

i-Alex (approved):

albertog78 (approved):

qa/rpc-tests/sc_rpc_cmds_fee_handling.py
alsala: done

PR #208, opened 2021-10-13
Added the new MaturityHeightIndex DB needed by the getblockexpanded RPC call

MarcoOl94:


src/main.cpp
```diff
@@ -5789,6 +5841,10 @@ bool InitBlockIndex() {
     fTxIndex = GetBoolArg("-txindex", false);
     pblocktree->WriteFlag("txindex", fTxIndex);
 
+    // Use the provided setting for -maturityheightindex in the new database
+    fMaturityHeightIndex = GetBoolArg("-maturityheightindex", false);
+    pblocktree->WriteFlag("maturityheightindex", fMaturityHeightIndex);
```
alsala, line 5846: It is correctly written, but a ReadFlag() is missing at the node statup (see the analogous pblocktree->ReadFlag("txindex", fTxIndex);)

src/maturityheightindex.h
```diff
@@ -0,0 +1,97 @@
...
+        return 4;
+    }
+    template<typename Stream>
+    void Serialize(Stream& s, int nType, int nVersion) const {
+        ser_writedata32be(s, blockHeight);
+    }
+    template<typename Stream>
+    void Unserialize(Stream& s, int nType, int nVersion) {
+        blockHeight = ser_readdata32be(s);
+    }
```
ptagl, line 16: Since we don't need a custom behavior, I would suggest using the serialization macro to generate functions.

```suggestion
    ADD_SERIALIZE_METHODS;

    template <typename Stream, typename Operation>
    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
        READWRITE(blockHeight);
    }
```

src/maturityheightindex.h
```diff
@@ -0,0 +1,97 @@
...
+    void Serialize(Stream& s, int nType, int nVersion) const {
+        // Heights are stored big-endian for key sorting in LevelDB
+        ser_writedata32be(s, blockHeight);
+        certId.Serialize(s, nType, nVersion);
+    }
+    template<typename Stream>
+    void Unserialize(Stream& s, int nType, int nVersion) {
+        blockHeight = ser_readdata32be(s);
+        certId.Unserialize(s, nType, nVersion);
+    }
```
ptagl, line 48: Since we don't need a custom behavior, I would suggest using the serialization macro to generate functions.

```suggestion
    ADD_SERIALIZE_METHODS;
    template <typename Stream, typename Operation>
    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
        READWRITE(blockHeight);
        READWRITE(certId);
    }
```

src/maturityheightindex.h
```diff
@@ -0,0 +1,97 @@
...
+
+    void SetNull() {
+        blockHeight = 0;
+        certId.SetNull();
+    }
+};
+
+//This is needed because the CLevelDBBatch.Write requires 2 arguments (key, value)
+struct CMaturityHeightValue {
+    unsigned int dummy;
```
ptagl, line 67: What about using a `char` type so that we can reduce the value size from 4 bytes to 1 byte?

src/maturityheightindex.h
```diff
@@ -0,0 +1,97 @@
...
+
+    template<typename Stream>
+    void Serialize(Stream& s, int nType, int nVersion) const {
+        ser_writedata32be(s, dummy);
+    }
+
+    template<typename Stream>
+    void Unserialize(Stream& s, int nType, int nVersion) {
+        dummy = ser_readdata32be(s);
+    }
```
ptagl, line 80: Since we don't need a custom behavior, I would suggest using the serialization macro to generate functions.

```suggestion
    ADD_SERIALIZE_METHODS;
    template <typename Stream, typename Operation>
    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
        READWRITE(dummy);
    }
```

src/rpc/blockchain.cpp
```diff
@@ -811,6 +812,178 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+            + HelpExampleRpc("getblockexpanded", "height")
+        );
+
+    LOCK(cs_main);
+
+    std::string strHash = params[0].get_str();
+
+    if (!fMaturityHeightIndex)
+    {
+        throw JSONRPCError(RPC_TYPE_ERROR, "txindex option not set: can not retrieve info");
```
ptagl, line 895:
```suggestion
        throw JSONRPCError(RPC_TYPE_ERROR, ", maturityHeightIndex option not set: can not retrieve info");
```

src/rpc/blockchain.cpp
```diff
@@ -811,6 +812,178 @@ UniValue getblock(const UniValue& params, bool fHelp)
     return blockToJSON(block, pblockindex, verbosity >= 2);
 }
 
+UniValue getblockexpanded(const UniValue& params, bool fHelp)
+{
+    if (fHelp || params.size() < 1 || params.size() > 2)
+        throw runtime_error(
+            "getblock \"hash|height\" ( verbose )\n"
+            "\nIf verbosity is 1, returns an Object with information about the block.\n"
```
ptagl, line 820: Can we mention that this command only works if the `-maturityHeightIndex` flag is enabled?

src/rpc/blockchain.cpp
```diff
@@ -811,6 +812,178 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+
+    if (fHavePruned && !(pblockindex->nStatus & BLOCK_HAVE_DATA) && pblockindex->nTx > 0)
+        throw JSONRPCError(RPC_INTERNAL_ERROR, "Block not available (pruned data)");
+
+    if(!ReadBlockFromDisk(block, pblockindex))
+        throw JSONRPCError(RPC_INTERNAL_ERROR, "Can't read block from disk");
+
+    UniValue blockJSON = blockToJSON(block, pblockindex, verbosity >= 2);
+    
+    //Add certificates that became maure with this block
```
ptagl, line 949: Typo

```suggestion
    //Add certificates that became mature with this block    
```

src/rpc/blockchain.cpp
```diff
@@ -811,6 +812,178 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+        UniValue matureCertificate(UniValue::VARR);
+
+        int height = pblockindex->nHeight;
+        if (pblocktree == NULL)
+        {
+            throw JSONRPCError(RPC_TYPE_ERROR, "DB not initialized: can not retrieve info");
+        }
+        std::vector<CMaturityHeightKey> matureCertificatesKeys;
+        pblocktree->ReadMaturityHeightIndex(height, matureCertificatesKeys);
+        BOOST_FOREACH(const CMaturityHeightKey& key, matureCertificatesKeys)
```
ptagl, line 961: I would use the "normal" C++11 **range-based for loop** rather than the `BOOST_FOREACH`.

qa/rpc-tests/getblockexpanded.py
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env python2
+# Copyright (c) 2014 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_node, connect_nodes_bi, assert_true, assert_false, get_epoch_data, \
+    swap_bytes, disconnect_nodes
```
ptagl, line 9: `disconnect_nodes` is unused.

```suggestion
    swap_bytes
```

qa/rpc-tests/getblockexpanded.py
```diff
@@ -0,0 +1,260 @@
...
+            "sc1", scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash, constant, [node1Addr], [bwt_amount2])
+
+        amount_cert_2 = [{"address": node1Addr, "amount": bwt_amount2}]
+
+        self.nodes[0].sc_send_certificate(scid, epoch_number, quality,
+            epoch_cum_tree_hash, proof, amount_cert_2, FT_SC_FEE, MBTR_SC_FEE, CERT_FEE)
+        self.sync_all()
+
+        #Mine a block
+        self.nodes[0].generate(1)[0]
```
ptagl, line 105: Accessing the element `[0]` seems useless.
```suggestion
        self.nodes[0].generate(1)
```

qa/rpc-tests/getblockexpanded.py
```diff
@@ -0,0 +1,260 @@
...
+                cert_3_json = cert
+        assert_true(cert_3_json != {})
+        tipHeight = self.nodes[0].getblockcount()
+
+        #Test that we require -maturityheightindex=1 to run the getblockexpanded
+        try:
+            self.nodes[1].getblockexpanded("640")
+            assert(False)
+        except JSONRPCException as e:
+            errorString = e.error['message']
```
ptagl, line 159: `errorString` is not used in this scope, should we print it in the following debug log?

qa/rpc-tests/getblockexpanded.py
```diff
@@ -0,0 +1,260 @@
...
+        assert_equal(len(rpcDataByHashVerbosity['matureCertificate']), 0)
+        assert_equal(len(rpcDataByHeightVerbosity['matureCertificate']), 0)
+
+        self.nodes[0].invalidateblock(fake_block)
+        self.nodes[1].invalidateblock(fake_block)
+
+        self.nodes[0].sc_send_certificate(scid, epoch_number, quality,
+            epoch_cum_tree_hash, proof, amount_cert_4, FT_SC_FEE, MBTR_SC_FEE, CERT_FEE)
+        self.nodes[0].generate(1)[0]
+        self.nodes[0].generate(1)
```
ptagl, line 226: Can't we just generate two blocks?

```suggestion
        self.nodes[0].generate(2)
```

qa/rpc-tests/getblockexpanded.py
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env python2
+# Copyright (c) 2014 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_node, connect_nodes_bi, assert_true, assert_false, get_epoch_data, \
+    swap_bytes, disconnect_nodes
+from test_framework.mc_test.mc_test import *
```
ptagl, line 10: It would be better to avoid importing unused symbols.

```suggestion
from test_framework.mc_test.mc_test import CertTestUtils, generate_random_field_element_hex
```

src/rpc/blockchain.cpp
```diff
@@ -811,6 +812,178 @@ UniValue getblock(const UniValue& params, bool fHelp)
     return blockToJSON(block, pblockindex, verbosity >= 2);
 }
 
+UniValue getblockexpanded(const UniValue& params, bool fHelp)
+{
+    if (fHelp || params.size() < 1 || params.size() > 2)
+        throw runtime_error(
+            "getblock \"hash|height\" ( verbose )\n"
```
alsala, line 819: should be 'getblockexpanded' (not 'getblock')

src/main.cpp
```diff
@@ -3702,8 +3743,10 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         }
 #endif // ENABLE_ADDRESS_INDEXING
 
-        if (fTxIndex)
+        if (fMaturityHeightIndex)
         {
+            //Remove the certificates from the MaturityHeight DB related to the ceased sidechains
+            view.HandleMaturityHeightIndexSidechainEvents(pindex->nHeight, pblocktree, maturityHeightValues);
             view.HandleTxIndexSidechainEvents(pindex->nHeight, pblocktree, vTxIndexValues);
```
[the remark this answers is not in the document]
MarcoOl94: Right, thanks!

src/txdb.cpp
```diff
@@ -448,6 +450,47 @@ bool CBlockTreeDB::WriteTxIndex(const std::vector<std::pair<uint256, CTxIndexVal
...
+        }
+    }
+
+    return true;
+}
+
+bool CBlockTreeDB::UpdateMaturityHeightIndex(const std::vector<std::pair<CMaturityHeightKey,CMaturityHeightValue>> &vect) {
+    CLevelDBBatch batch;
+    for (std::vector<std::pair<CMaturityHeightKey,CMaturityHeightValue> >::const_iterator it=vect.begin(); it!=vect.end(); it++)
+        if (it->second.IsNull()) {
```
rdoria1, line 486: Add comment to specify this is convention we decided -> `If the value is null we mean we want to erase the pair from the DB otherwise we persist it`

src/txdb.cpp
```diff
@@ -448,6 +450,47 @@ bool CBlockTreeDB::WriteTxIndex(const std::vector<std::pair<uint256, CTxIndexVal
...
+
+    CDataStream ssKeySet(SER_DISK, CLIENT_VERSION);
+    ssKeySet << make_pair(DB_MATURITY_HEIGHT, CMaturityHeightIteratorKey(height));
+    pcursor->Seek(ssKeySet.str());
+
+    while (pcursor->Valid()) {
+        boost::this_thread::interruption_point();
+        try {
+            leveldb::Slice slKey = pcursor->key();
+            CDataStream ssKey(slKey.data(), slKey.data()+slKey.size(), SER_DISK, CLIENT_VERSION);
```
rdoria1, line 464: If performance becomes an issue,

We could try to persist
- key: height
- value: List of mature certificates at heigh specified in key

Not sure how Slice works and if this is as performant as the solution proposed.

src/rpc/blockchain.cpp
```diff
@@ -811,6 +812,180 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+        std::vector<CMaturityHeightKey> matureCertificatesKeys;
+        pblocktree->ReadMaturityHeightIndex(height, matureCertificatesKeys);
+        for (const CMaturityHeightKey& key : matureCertificatesKeys) 
+        {
+            if (verbosity == 2)
+            {
+                UniValue objCert(UniValue::VOBJ);
+                CScCertificate certAttempt;
+                uint256 hashBlock{};    
+                if (GetCertificate(key.certId, certAttempt, hashBlock, true))
```
albertog78, line 969: Instead of allowSlow=true wouldn't be better to require enabling the txIndex when maturityheightindex is on?

src/txdb.cpp
albertog78: The drawback of this approach is that several updates are going to be performed on the same object implying deserializing, serializing and saving the whole list each time (consider that epoch lengths could be different between sidechains and different blocks can trigger the modification of the same maturityheight).

MarcoOl94: alberto's comment is related to the rdoria's comment.

PR #214, opened 2021-10-20
Modification for ws get_top_quality_certificate for not relying on tx…

alsala:
…index options being enabled on zend

### DIFF
--- a/src/zen/websocket_server.cpp
+++ b/src/zen/websocket_server.cpp
@@ -545,7 +545,6 @@ private:
                 mempoolTopQualityCert.push_back(Pair("quality", topQualCert.quality));
                 mempoolTopQualityCert.push_back(Pair("epoch", topQualCert.epochNumber));
                 mempoolTopQualityCert.push_back(Pair("certHash", topQualCertHash.GetHex()));
-                mempoolTopQualityCert.push_back(Pair("rawCertificateHex", certHex));
                 mempoolTopQualityCert.push_back(Pair("fee", FormatMoney(certFee)));
             }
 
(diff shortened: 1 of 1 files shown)


albertog78 (approved):

src/zen/websocket_server.cpp
```diff
@@ -556,18 +556,19 @@ class WsHandler
                 CScCertificate topQualCert;
                 uint256 blockHash;
 
+                chainTopQualityCert.push_back(Pair("quality", topQualityCertQuality));
+                chainTopQualityCert.push_back(Pair("certHash", sidechainInfo.lastTopQualityCertHash.GetHex()));
+                chainTopQualityCert.push_back(Pair("epoch", sidechainInfo.lastTopQualityCertReferencedEpoch));
+
                 if (GetCertificate(sidechainInfo.lastTopQualityCertHash, topQualCert, blockHash, true)) {
```
albertog78, line 563: Passing allowSlow = true implies loading and parsing the full block from disk. Considering that now the sidechain cannot expect to always have the rawCertificateHex returned (eg. it will not be returned if the cert had no bt outputs and change was already spent) why not always avoiding this expensive operation?

albertog78 (changes requested):

src/zen/websocket_server.cpp
alsala: After discussion it has been decided to remove the call to GetCertificate()

i-Alex (approved):

albertog78 (approved):

PR #216, opened 2021-10-21
Zendoo macos compatibility

rdoria1:
- Pull in patch fixing OpenSSL build error on older MacOS
- Remove --enable-werror flag

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -183,7 +183,7 @@ os: linux
 
 env:
   global:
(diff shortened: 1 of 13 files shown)


cronicc (approved):
Approving and merging as we need the addressindexing build script changes in development. Remaining things to be fixed are split into new issue https://github.com/HorizenOfficial/zend_oo/issues/220

